PR #6, opened 2021-03-14
Add "version" key

ste94pz:
Home Assistant gives an error if a custom integration doesn't include the "version" key in the manifest.json


anotherjulien:
Thank you!

PR #29, opened 2021-10-19
Update it.json

xtimmy86x:


anotherjulien:
Thank you!
Had I known you would contribute, I would have waited a couple of hours before publishing the latest release :-D

PR #40, opened 2022-01-05
Update config_flow.py

xtimmy86x:
Update ssdp discovery code for working with new core update


anotherjulien (approved):
Thank you VERY much for your help!

PR #50, opened 2022-05-20
Fix format_mac when registering a device

jeroen-vankelecom:
```
  File "/usr/src/homeassistant/homeassistant/helpers/device_registry.py", line 134, in format_mac
    if len(to_test) == 17 and to_test.count(":") == 5:
TypeError: object of type 'MACAddress' has no len()
```

The `format_mac` function expects an argument of type str, but receives `MacAddress` class. Apperently, both this MyHome integration and HomeAssistant do a check on the MacAddres/serial number. Did not want to change this.


anotherjulien:
Thank you Jeroen!

PR #55, opened 2022-07-21
Cover window class

AlessioBazzanella:
Added `class` option in Covers to distinguish between  `shutter` and `window`


anotherjulien:
Thank you for your contribution!
I'll admit I did not know one could use the shutter modules to open and close windows.
Didn't the "Display as" mechanism introduced in HA Core a few months ago work for that?

PR #65, opened 2022-11-17
Fixed HVACAction modes to correct HA core naming

IAmVespa:
Change this avoid the integration to crash any time the HVAC mode moves to heating


anotherjulien:
Thank you for your contribution!

PR #109, opened 2024-02-08
Improve logging to ease troubleshooting

f18m:
Hi @anotherjulien,

Thanks for this HA custom component, it's really useful.
I found some trouble in setting it up (although now it works well for me).
To find out what was wrong I introduce a few logging statements in the code. 

So here it is a PR which I think will help newcomers to use/troubleshoot trivial issues:

Changelog:
* Add logs at either DEBUG, INFO or ERROR level to explain why the 'async_setup_entry' function marks a myHome gateway as "config not ready" 
* Use consistently Unix line ends on all source code
* Fix for https://github.com/anotherjulien/MyHOME/issues/108: see line 441 of gateway.py

About the EOLs: I noticed that some files in this repository use Windows-style EOLs (\r\n) while most of the rest is using Linux-style EOLs (\n):

```
$ cd MyHOME/custom_components/myhome
$ grep -rl $'\r' *.py *.json
config_flow.py
const.py
__init__.py
manifest.json
```

Having inconsistent EOL encoding is confusing editors, specially on embedded systems. 
So I think it would be good to stay consistent (most of people use the default git setting 'core.autocrlf' set to false which means that git will not do any EOL conversion on the fly at checkout/commit time)

Final point: as I developed these changes using the dev-container of HA (https://developers.home-assistant.io/docs/development_environment/) when I saved the source code it has been reformatted using "ruff" utility... that's why in the diff you see so many long lines broken into shorter lines. 
I hope this is not a problem.


GreenGrassBlueOcean:
Thank you for this contribution. As announced in https://github.com/anotherjulien/MyHOME/issues/229, the integration has moved to the OpenWebNet-HA organization and is undergoing a comprehensive v2 reboot. We are closing legacy v1 pull requests to establish a clean baseline. Please feel free to test against the new v2 architecture and submit updated PRs once the core reboot lands.

PR #131, opened 2024-06-10
Update gateway.py, added a test that dimmer exists in configuration b…

BWsyl:
…efore running code

I was testing my installation and encountered a case that crashed the myhome integration.
I a dimmer is not defined in the configuration, and you press a button triggering a dimmer light, it made the integration crashed.

I added a test to check that the dimmer light really exists before replicating the received order from the gateway.


gPinato:
I think the bug you encountered is exactly what I experienced here
https://github.com/anotherjulien/MyHOME/issues/125

PR #217, opened 2026-07-01
Fix non-string manufacturer passed to device registry

G3z:
The manual config flow wrapped gateway metadata values in single-element tuples (a stray trailing comma), so `manufacturer` was stored as a tuple instead of a string. Home Assistant's device registry rejects non-string values and warns this will stop working in 2026.12.0.

- config_flow.py: store the manual-entry values as plain strings/None.
- __init__.py: defensively coerce list/tuple values to a string when building the gateway device entry, so existing config entries that already stored the tuple no longer trigger the warning.

### DIFF
--- a/custom_components/myhome/__init__.py
+++ b/custom_components/myhome/__init__.py
@@ -31,6 +31,17 @@ CONFIG_SCHEMA = cv.config_entry_only_config_schema(DOMAIN)
 PLATFORMS = ["light", "switch", "cover", "climate", "binary_sensor", "sensor"]
 
 
+def _coerce_str(value):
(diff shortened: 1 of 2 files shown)


GreenGrassBlueOcean:
Thank you for this contribution. As announced in https://github.com/anotherjulien/MyHOME/issues/229, the integration has moved to the OpenWebNet-HA organization and is undergoing a comprehensive v2 reboot. We are closing legacy v1 pull requests to establish a clean baseline. Please feel free to test against the new v2 architecture and submit updated PRs once the core reboot lands.

PR #313, opened 2026-09-13
Added attributes for the thermostat (knob position)

DinoMesina:
Added attributes for the thermostat

Page 17 of https://developer.legrand.com/uploads/2019/12/WHO_4.pdf

FIXED tests

### DIFF
--- a/custom_components/myhome/climate.py
+++ b/custom_components/myhome/climate.py
@@ -33,6 +33,12 @@ from OWNd.message import (
     MESSAGE_TYPE_MODE,
     MESSAGE_TYPE_MODE_TARGET,
     MESSAGE_TYPE_TARGET_TEMPERATURE,
+    LOCAL_CONTROL_NORMAL,
+    LOCAL_CONTROL_OFFSET,
+    LOCAL_CONTROL_OFF,
+    LOCAL_CONTROL_PROTECTION,
+    LOCAL_CONTROL_OVERRIDE,
+    LOCAL_CONTROL_UNKNOWN,
     OWNHeatingCommand,
     OWNHeatingEvent,
 )
@@ -434,6 +440,7 @@ class MyHOMEClimate(MyHOMEEntity, ClimateEntity):
         self._attr_current_humidity = None
         self._target_temperature = None
         self._local_offset = 0
+        self._knob_pos = "UNKNOWN"
         self._local_target_temperature = None
 
         self._attr_hvac_mode = None
@@ -444,6 +451,8 @@ class MyHOMEClimate(MyHOMEEntity, ClimateEntity):
         """Return device specific attributes."""
         attrs = {
             "local_offset": self._local_offset,
+            "local_target_temperature": self._local_target_temperature,
+            "knob_pos":self._knob_pos,
         }
         if self._fan:
             attrs["fan_mode"] = self._attr_fan_mode
@@ -690,6 +699,19 @@ class MyHOMEClimate(MyHOMEEntity, ClimateEntity):
                 message.human_readable_log,
             )
             self._local_offset = message.local_offset
+            if message.local_control_state is None:
+                self._knob_pos = "UNKNOWN"
+            else:
+                if message.local_control_state == LOCAL_CONTROL_NORMAL or message.local_control_state == LOCAL_CONTROL_OFFSET:
+                    self._knob_pos = f"{message.local_offset:+d}" if message.local_offset != 0 else "0" 
+                elif message.local_control_state == LOCAL_CONTROL_OFF:
+                    self._knob_pos = "OFF"
+                elif message.local_control_state == LOCAL_CONTROL_PROTECTION:
+                    self._knob_pos = "*"
+                elif message.local_control_state == LOCAL_CONTROL_OVERRIDE:
+                    self._knob_pos = "?"
+                else:
+                    self._knob_pos = "UNKNOWN"
             if self._target_temperature is not None:
                 self._local_target_temperature = self._target_temperature + self._local_offset
         elif message.message_type == MESSAGE_TYPE_LOCAL_TARGET_TEMPERATURE:


GreenGrassBlueOcean (changes requested):
Thank you for this PR @DinoMesina! Supporting physical thermostat knob positions via WHO=4 Dimension 13 (as documented on page 17 of `WHO_4.pdf`) is a very welcome addition.

During local testing against our test suite and protocol verification tools, we identified a critical runtime crash that needs to be addressed before merging:

---

### 🔴 Critical Runtime Bug: `TypeError` on OFF / Protection / Override States

In `OWNd.message.OWNHeatingEvent`, whenever a thermostat reports a local control state of OFF (`*#4*where*13*4##`), Protection (`*#4*where*13*5##`), Override (`*#4*where*13*6##`), or Unknown, **`message.local_offset` is `None`** (not `0` or an integer).

Because `self._local_offset` is assigned directly:
```python
self._local_offset = message.local_offset  # becomes None
...
if self._target_temperature is not None:
    self._local_target_temperature = self._target_temperature + self._local_offset
```
1. **Crash in `handle_event`**: If the zone already has a target temperature set (which is normal operation for active thermostats), evaluating `self._target_temperature + self._local_offset` raises:
   ```text
   TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'
   ```
   This unhandled exception crashes the event handler.
2. **Crash in `async_set_temperature`**: When `self._local_offset` is `None`, any subsequent temperature change from the HA frontend or automations executes:
   ```python
   target_temperature = kwargs.get("temperature", self._local_target_temperature) - self._local_offset
   ```
   which also crashes with `TypeError`.

#### Why tests previously passed
In `test_climate_knob_positions_coverage`:
- `mock_event.local_offset` was hardcoded to `0` across all test states (including OFF and PROTECTION where real OWNd messages return `None`).
- `entity._target_temperature` was `None`, so the addition was never executed during the test run.

---

### 🛠️ Suggested Fix

In `custom_components/myhome/climate.py`:
1. Safeguard `_local_offset` to ensure it always remains a valid integer (defaulting to `0` when `None`):
```python
        elif message.message_type == MESSAGE_TYPE_LOCAL_OFFSET:
            LOGGER.debug(
                "%s %s",
                self._gateway_handler.log_id,
                message.human_readable_log,
            )
            self._local_offset = message.local_offset if message.local_offset is not None else 0
            if message.local_control_state in (LOCAL_CONTROL_UNKNOWN, None):
                self._knob_pos = "UNKNOWN"
            elif message.local_control_state == LOCAL_CONTROL_OFFSET:
                self._knob_pos = f"{self._local_offset:+d}"
            elif message.local_control_state == LOCAL_CONTROL_NORMAL:
                self._knob_pos = "0"
            elif message.local_control_state == LOCAL_CONTROL_OFF:
                self._knob_pos = "OFF"
            elif message.local_control_state == LOCAL_CONTROL_PROTECTION:
                self._knob_pos = "*"
            elif message.local_control_state == LOCAL_CONTROL_OVERRIDE:
                self._knob_pos = "?"
            else:
                self._knob_pos = "UNKNOWN"

            if self._target_temperature is not None:
                self._local_target_temperature = (
                    self._target_temperature + self._local_offset
                )
```
2. Formatting: Add a space after the colon in `"knob_pos": self._knob_pos` in `extra_state_attributes`.
3. In `tests/test_component_climate.py`, add a test case using real OWNd events with an initialized target temperature:
```python
    # Test real frame with OFF state (where OWNd sets local_offset to None)
    entity._target_temperature = 20.0
    event_off = OWNHeatingEvent("*#4*1*13*4##")
    entity.handle_event(event_off)
    assert entity._knob_pos == "OFF"
    assert entity._local_offset == 0
    assert entity._local_target_temperature == 20.0
```

---

### 💡 Tip: Protocol & Frame Verification with `openwebnet-mcp`

For protocol-level development, we actively recommend testing and cross-referencing OpenWebNet frames against the **`openwebnet-mcp`** MCP server.

`openwebnet-mcp` provides semantic grammar validation, WHO dimension lookups, and protocol conformance checks directly in local development environments and AI-assisted pair-programming sessions. Running frames like `*#4*WHERE*13*OL##` through `parse_and_validate_frame` and checking dimension behavior against `get_who_spec` helps catch protocol nuances, edge cases, and parser return types (like `None` vs `0`) before hitting production bus hardware or Home Assistant runtime environments.

DinoMesina:
Maybe I have fixed the issues.

Thanks :-)

GreenGrassBlueOcean:
Thanks @DinoMesina, the `None` safeguard and the real-frame test look good! 👍

Just one thing left, and it's not in your code: the three **HA Compatibility** jobs fail on a snapshot test, because the climate entity now exposes two new attributes that the stored snapshot doesn't know about yet:

```text
tests/test_legacy_platforms.py::test_legacy_platforms_setup_and_execution
E         +     'knob_pos': 'UNKNOWN',
E               'local_offset': 0,
E         +     'local_target_temperature': None,
```

(The "Run Pytest Suite & Coverage" job is green only because it runs with `--snapshot-update`, so it doesn't catch this.)

### Fix

Either regenerate the snapshot locally:

```bash
pytest tests/test_legacy_platforms.py --snapshot-update
```

…or edit `tests/snapshots/test_legacy_platforms.ambr` by hand in the `climate_state` block (attributes are sorted alphabetically):

```diff
       'hvac_modes': list([
         <HVACMode.OFF: 'off'>,
         <HVACMode.AUTO: 'auto'>,
         <HVACMode.HEAT: 'heat'>,
         <HVACMode.COOL: 'cool'>,
       ]),
+      'knob_pos': 'UNKNOWN',
       'local_offset': 0,
+      'local_target_temperature': None,
       'max_temp': 40,
```

Then commit the `.ambr` file together with your change.

### Optional nit (not blocking)

In `custom_components/myhome/climate.py`, inside `handle_event`, the `MESSAGE_TYPE_LOCAL_OFFSET` branch now does two things with the offset:

1. The first line stores a safe value in `self._local_offset`, replacing `None` with `0`.
2. A few lines later, the `LOCAL_CONTROL_OFFSET` case builds the knob string, but it reads the **raw** `message.local_offset` again instead of the value you just made safe.

```diff
             self._local_offset = message.local_offset if message.local_offset is not None else 0
             ...
             elif message.local_control_state == LOCAL_CONTROL_OFFSET:
-                self._knob_pos = f"{message.local_offset:+d}"
+                self._knob_pos = f"{self._local_offset:+d}"
```

This is not a bug today: when the state is `LOCAL_CONTROL_OFFSET`, OWNd always provides an integer offset, so `message.local_offset` is never `None` there. It is just cleaner for all the code in that branch to use the one safe variable, so nobody later has to reason about which of the two can be `None`.

Once the snapshot is in, this should be good to merge. Thanks again!

DinoMesina:
Thank you so much for helping me :-)

GreenGrassBlueOcean (approved):
All checks passing and snapshots updated. Thanks @DinoMesina!

Issue #247, opened 2026-09-09
fix: Resolve dry contact ghost devices and zero-padded WHERE sensor updates (WHO 1/25)


GreenGrassBlueOcean:
### Problem Description

Following live testing on real Legrand hardware reported by @anotherjulien in [#232 (comment)](https://github.com/OpenWebNet-HA/MyHOME/pull/232#issuecomment-5606151001):

1. **Dry Contact Device Triplication & Ghost Devices (WHO = 25)**:
   - When migrating existing dry contact devices (e.g. Legrand 3477 garage door contact), Home Assistant created:
     1. The active device with the correct entity.
     2. A duplicate legacy device with the same name containing 0 entities (caused by the device identifier changing from `{mac}-{device_id}` in v1 to `{mac}-25-{device_id}` in v2 without pruning empty devices from `device_registry`).
     3. A third device `"Dry contact garage door"` containing an `unavailable` entity (caused by unique ID mismatch during registry restoration where `entry.unique_id` was not matched by `MyHOMEDryContact` or orphaned secondary entities).

2. **Motion / Illuminance Sensor Unresponsive Due to Zero-Padded Addresses (WHO = 1)**:
   - Physical Legrand 048834 light/motion sensors transmit 4-digit zero-padded point-to-point OpenWebNet frames:
     ```
     [RX] *#1*0015*6*33338##   (Illuminance value: 33,338 lux)
     [TX] *#1*0015*5##         (PIR sensitivity query)
     [RX] *#1*0015*5*2##       (PIR sensitivity reply: High)
     [TX] *#1*0015*7##         (Motion timeout query)
     [RX] *#1*0015*7*0*0*10##  (Motion timeout reply: 10s)
     ```
   - In `myhome.yaml` or registry, addresses are configured unpadded (`where: 15`).
   - Because `_sensor_address` and `_handle_binary_sensor_message` did not normalize zero-padded addresses (`0015` vs `15`), frames for `0015` never matched entities keyed under `15`.
   - Result: Illuminance sensor stayed `unknown` and motion binary sensor stayed `unavailable`.

---

### Proposed Solution

1. **`normalize_where` Helper**:
   - Canonicalize SCS numeric addresses (e.g. `0015` -> `15`, `01` -> `1`, `0015#4#01` -> `15#4#01`) across `binary_sensor.py` and `sensor.py`.
   - Dispatch and index messages under both raw and normalized addresses.

2. **Empty Device Cleanup**:
[...]

PR #316, opened 2026-09-14
fix: synthetic plant fixtures, privacy at the source, CI on the core users run (Python 3.14 / 2026.9)

GreenGrassBlueOcean:
## Summary

@Interstellar0verdrive and @TheDarkWizard were right about the plant fixtures, and 96d86a8 removed them. That also removed the trace-replay coverage: `tests/test_trace_replay.py` requires at least one plant, and the #247 golden tests read its `myhome.yaml`, so the base branch is red right now. This PR puts the coverage back **without the data**.

What the fixtures contained: room and family names in `myhome.yaml`, the gateway's LAN address and MAC, the config-entry id, one `password` value, the time zone, and the reporters' handles as directory names. The replay harness needs none of that — only addresses, platforms, options and the frames.

### What changes

- **`scripts/anonymize_plant_fixture.py`** rewrites a fixture **in place, line by line**:
  - devices become `<platform>_<where>` / `<Label> <where>` (`light_10` / `Light 10`): the address is the name. The class stays in the *key* (`sensor_51_power`, `sensor_51_energy`) so a meter's two entries keep one device name and Home Assistant derives the ids it always did (`sensor.sensor_51_power`); a bus interface stays part of the identity (`cover_11i02` / `Cover 11I02`); room-name section comments become `# ── group n ──`;
  - IPs → `192.0.2.0/24` (RFC 5737), MACs → `00:03:50:00:<issue number>` (the OUI is BTicino's public one), entry id → a synthetic ULID-shaped string, `password` → `null`, time zone → `UTC`;
  - every other byte — frames, quoting, indentation, file headers — is kept, so a diff against an original shows only the identity lines (the #247 `myhome.yaml`: 206 of 703 lines; each diagnostics file: 4–7 lines; frames stay one per line). Same input, same output.
- **Four fixtures generated from the originals** with it: `issue_247_myhomeserver1`, `issue_292_myhomeserver1`, `issue_297_f454`, `mh200_physical_plant` — named after the issue and the gateway model, not the reporter.
- **`tests/test_fixture_privacy.py`** fails the build if a committed fixture contains a LAN address, a non-synthetic MAC, a secret or a non-generic device name, and covers the script.
- Tests reading the fixtures use the new entity ids; the golden-corpus credit and the ROADMAP table point at the new directories; README documents the rule for contributed traces.

### What this PR cannot do

The originals remain in this branch's **history** (commits `079721d`, `f69e2cf`, `4a48f81`, `11e1df9`, and 96d86a8's deletion diff itself) and in every fork. Removing them for good means rewriting `v2-phase1-architecture`, force-pushing, asking GitHub support to purge cached views, and every contributor with a branch on it (#313 among them) rebasing. That is a maintainers' decision; the tree is clean either way.

## Test plan

- [x] `pytest tests` on core 2026.9.2: the fixture-driven suites pass (`test_trace_replay`, `test_golden_conformance`, the #247 golden tests) plus 6 new tests; the remaining failures on `v2-phase1-architecture` are the core-2026.9 incompatibilities fixed in #323, not this change
- [x] Fixtures reproducible: running the script on the original files yields the committed fixtures byte for byte
- [x] `grep` of the tree for the reporters' handles, the real MACs / IPs and the family names: nothing
- [x] CI


Interstellar0verdrive (changes requested):
Thanks, this is the right shape: generic names, a reproducible script and a test that fails the build on a regression. I scanned the four new fixtures for names, room words (Italian and Dutch), LAN addresses, real MACs, passwords and time zones: all clean.

Two leftovers before I approve:

1. `tests/fixtures/plants/mh200_physical_plant/diagnostic_summary.json`, line 28: a Windows `file_path` still carries a local user name. It's yours, not a reporter's, but the rule should hold for everyone, and `test_fixture_privacy.py` doesn't catch paths.
2. `issue_247_myhomeserver1/diagnostic_summary.json` still lists the 56 custom integrations installed in that house (energy supplier included). It's the indirect exposure discussed in #229: it tells a stranger what that household runs. The replay harness doesn't need it; reduce `custom_components` to `myhome` alone (or drop the key) and let the privacy test cover that field too.

On the history: agreed that rewriting the branch is a separate decision for the maintainers; the tree being clean is the first step.

Interstellar0verdrive (changes requested):
Updating my review state: the earlier requests are closed, both done and verified at `e236549`. The open request is the split into three PRs described in my comment above.

Interstellar0verdrive (approved):
Approving: the split is exactly what I asked for, and at `4e823b2` this PR is only the fixtures, the anonymizer, the privacy test and the docs that go with them. I re-checked the four plants: documentation-range IPs, synthetic MACs, no local paths, every device named after its address, `custom_components` down to `myhome`. CI is green and @xtimmy86x has already signed off on the sanitization.

On the red base: that's on our side, not yours. Removing the fixtures folder on `v2-phase1-architecture` was deliberate (clean tree first), and this merge is what makes the base green again. Merging now, in the order you listed; #334 should go green on its own right after.

Thanks for the extra work the split took. Noted for next time: one review round per PR, all points at once.

Issue #311, opened 2026-09-12
Need help configuring Alarm feature (F454)


gimprota76:
Hello,

First of all, thank you for the great work on this integration.

I recently enabled the new Alarm feature with my BTicino F454 alarm system, but I am not sure if I have configured it correctly.

Home Assistant creates 9 alarm entities (Alarm 0 to Alarm 8). I can see their status, but I do not understand which one should be considered the main alarm panel.

I also have a few issues:

- I cannot arm or disarm the alarm from Home Assistant.
- When I arm or disarm the alarm using my transponder/key, the entity states do not seem to update correctly.
- I am not sure how the different alarm entities are supposed to work.

My goal is to have a simple configuration with:
- One alarm card/panel.
- Correct alarm status updates.
- Optional arm/disarm control from Home Assistant, if supported.

Could you please provide a simple configuration example or explain how the new Alarm feature should be configured with an F454 system?

Thank you very much for your support. I am available to perform tests and provide logs if needed.

<img width="717" height="796" alt="Image" src="https://github.com/user-attachments/assets/a9fd1da6-2b28-4938-9e4d-ea8b30e0a086" />

Best regards,
Giovanni

GreenGrassBlueOcean:
Hello @gimprota76,

Thank you for reporting this and for using the integration!

Here is an explanation of how the Alarm feature works with your F454 and how to configure a clean single-panel setup:

---

### 1. Which entity is the main alarm panel?
* **`Alarm 0` (`alarm_control_panel.alarm_0`) is your main panel.**
* In the OpenWebNet protocol (`WHO = 5` — Burglar Alarm / *Antifurto*):
  * **`WHERE = 0`** represents the **Central Unit** (*Centrale Antifurto*).
  * **`WHERE = 1` through `8`** represent the **8 individual zones / partitions** (*Settori*) managed by the central panel (e.g. 3486 / 3485).
* The integration automatically discovers and creates an `alarm_control_panel` entity for every address that sends or receives a WHO 5 frame on the SCS bus. Because your central unit reports status for both itself (`0`) and its 8 partitions (`1`–`8`), Home Assistant generated 9 entities (`Alarm 0` to `Alarm 8`).

---

### 2. Why don't states update when using your transponder / key?
[...]

Issue #310, opened 2026-09-12
Deprecated via_device argument in gateway.py will stop working in Home Assistant 2027.8


xtimmy86x:
### Description

Home Assistant reports that MyHOME calls `device_registry.async_get_or_create()` with the deprecated `via_device` argument.

The warning identifies `custom_components/myhome/gateway.py`, line 114, and states that this usage will stop working in Home Assistant `2027.8.0`.

### Environment

* Home Assistant version: 2026.9.1
* MyHOME version '2.0.0b12'
* OWNd version: `2.0.0b6`
* Gateway: BTicino F454

### Observed behavior

The following warning appeared at `2026-09-11 08:50:44`:

```text
Detected that custom integration 'myhome' calls
`device_registry.async_get_or_create` with a deprecated `via_device`
parameter; use `via_device_id` instead at
custom_components/myhome/gateway.py, line 114:
device_registry.async_get_or_create(.

This will stop working in Home Assistant 2027.8.0,
please create a bug report at
https://github.com/OpenWebNet-HA/MyHOME/issues
```

The log does not identify which specific device triggered this registration call.

### Reproduction

The warning was observed while running MyHOME with an F454 gateway.

To reproduce, exercise the device registration path in `gateway.py` on a Home Assistant version that reports this deprecation. The exact device/discovery event that triggered it in this session is not available in the log.

### Expected behavior

MyHOME should register device relationships through the supported device registry API without deprecation warnings.

### Impact

The log shows a compatibility warning rather than an immediate registration failure. However, it explicitly identifies a future breaking change in Home Assistant `2027.8.0`.

### Suggested investigation / fix

Migrate the parent gateway relationship to `via_device_id`.

This should include resolving the gateway's device registry entry and passing its registry ID. It should not be treated as a simple argument rename with the previous value left unchanged.

Please also check:

* Other device registration calls that use `via_device`.
* Compatibility with the integration's minimum supported Home Assistant version.
* Registration order, so the parent gateway exists before child devices reference it.
* Multi-gateway behavior, so each child remains linked to the correct gateway.

### Validation

* Register or rediscover devices and verify that the warning disappears.
* Confirm that child devices still show the correct gateway relationship.
* Reload/restart and verify that device identities and relationships remain stable.
* Check that the migration does not create duplicate devices.

Interstellar0verdrive:
A note for whoever picks this up on the v2 side (the report itself is complete, nothing to add to it). 

[...]

Issue #309, opened 2026-09-12
Blocking filesystem operations in get_ownd_version() during integration setup


xtimmy86x:
### Description

During MyHOME startup, Home Assistant reports blocking filesystem operations in the event loop when the integration retrieves the installed OWNd version.

Three warnings are generated: `listdir`, `read_text`, and `open`. They all originate from the same call to `importlib.metadata.version("OWNd")`, so I am reporting them together.

### Environment

* Home Assistant version: 2026.9.1
* MyHOME version 2.0.0b12
* OWNd version: `2.0.0b6`
* Gateway: BTicino F454

### Steps to reproduce

1. Start Home Assistant with the MyHOME integration configured.
2. Allow the integration to set up.
3. Check the Home Assistant logs for blocking-call warnings.

This was observed during the startup captured in the attached log.

### Actual behavior

At `2026-09-11 08:21:01`, Home Assistant reports:

```text
Detected blocking call to listdir with args
('/config/deps/lib/python3.14/site-packages',)
inside the event loop by custom integration 'myhome'
at custom_components/myhome/const.py, line 18:
return importlib.metadata.version("OWNd")
```

```text
Detected blocking call to read_text with args
(PosixPath('/usr/local/lib/python3.14/site-packages/ownd-2.0.0b5.dist-info/METADATA'),)
inside the event loop by custom integration 'myhome'
at custom_components/myhome/const.py, line 18:
return importlib.metadata.version("OWNd")
```

```text
Detected blocking call to open with args
(PosixPath('/usr/local/lib/python3.14/site-packages/ownd-2.0.0b5.dist-info/METADATA'),)
inside the event loop by custom integration 'myhome'
at custom_components/myhome/const.py, line 18:
return importlib.metadata.version("OWNd")
```

The relevant application frames are:

```text
File "/config/custom_components/myhome/__init__.py", line 293, in async_setup_entry
    get_ownd_version(),

File "/config/custom_components/myhome/const.py", line 18, in get_ownd_version
    return importlib.metadata.version("OWNd")
```

### Expected behavior

Retrieving dependency version information should not perform blocking filesystem operations in Home Assistant's event loop.

### Impact

This is a confirmed blocking-I/O warning during setup. The log does not establish a user-visible freeze or a setup failure, but the filesystem work can delay other event-loop tasks while it executes.

### Suggested investigation / fix

Run the complete version lookup in Home Assistant's executor, for example:

```python
[...]

Issue #308, opened 2026-09-12
[Bug] F454 + 3455/L4577: recurring NACKs from temperature polling; readings continue when the request is disabled


xtimmy86x:
## Summary

MyHOME repeatedly sends this request for my external temperature sensor:

```text
*#4*100*15##
```

The original Home Assistant log contains **43 error entries** for this request, recurring approximately every five minutes during the final part of the log.

However, the temperature entity receives updates and its history changes throughout the day.

I performed a temporary local test to disable both periodic polling and the initial request for this sensor. In the resulting trace, the specific TX request disappears, while incoming temperature frames continue at approximately **8.6-second intervals**, including a temperature change.

This suggests that the polling strategy needs review for this device combination. The test does not establish that dimension 15 requests are universally unsupported.

## Environment

Environment reported by the September 12 diagnostic bundles:

| Component | Version / configuration |
|---|---|
| Home Assistant | 2026.9.1 |
| MyHOME integration | 2.0.0b12 |
| OWNd protocol engine | 2.0.0b6 |
| Gateway | BTicino F454 |
| Gateway firmware | 2.0 |
| Connection | Ethernet TCP, port 20000 |
| Queue pacing | 0.05 seconds |
| Worker count | 1 |
| External temperature sensor | BTicino 3455 |
| Radio receiving interface | BTicino L4577 |
| Observed OpenWebNet address | `100` |

**Version distinction:** the original September 11 error log identifies OWNd **2.0.0b5**. The subsequent diagnostic traces and local tests were collected with **2.0.0b6**.

## Original symptoms

Representative entries from September 11, with the gateway IP redacted:

```text
2026-09-11 09:46:10.859 ERROR (MainThread) [custom_components.myhome]
[F454 gateway - <gateway-ip>] Could not send message `*#4*100*15##`. Retrying (1)...

2026-09-11 19:41:11.815 ERROR (MainThread) [custom_components.myhome]
[F454 gateway - <gateway-ip>] Could not send message `*#4*100*15##`. Retrying (1)...

2026-09-11 19:46:11.840 ERROR (MainThread) [custom_components.myhome]
[F454 gateway - <gateway-ip>] Could not send message `*#4*100*15##`. Retrying (1)...

2026-09-11 21:56:10.961 ERROR (MainThread) [custom_components.myhome]
[F454 gateway - <gateway-ip>] Could not send message `*#4*100*15##`. Retrying (1)...
```

[...]

PR #317, opened 2026-09-14
fix: HA quality-scale alignment + issue batch (#304 #305 #307 #308 #309 #310)

GreenGrassBlueOcean:
## Summary

A senior-review pass on the integration's HA core surface plus the fixes for six of the seven issues opened against 2.0.0b12 this week. Every change is covered; the suite is at **1335 passed, 100.0 % statement coverage**, `scripts/verify_ha_standards.py` passes, ruff is clean. **Live-tested on HA OS 18.2 / Core 2026.9.2 with an MH200N** (see Verification).

### Issues fixed

| Issue | Fix |
|---|---|
| #304 MH201 `Data received is not a message: None` | OWNd's `get_next()` returns `None` for the cycle in which it reconnects (documented contract). The listener now skips that cycle at DEBUG instead of warning and dispatching `None`. The accompanying `Event connection lost, reconnecting...` line is OWNd's own reconnect log for the gateway-side idle close. |
| #305 Bus-monitor timestamps 2 h behind | The card sliced the UTC ISO string. Frames now render through the browser's local zone (`_formatFrameTime`), matching the logbook. |
| #307 DALI DT8 HS ↔ CT toggling (part 1) | Colour-mode auto-promotion is additive: HS and COLOR_TEMP coexist, `color_mode` follows the last frame, and restore keeps the full set. The "lock supported features / reset stored gateway defaults" request from the same issue is **not** in this PR — see follow-ups. |
| #308 `*#4*100*15##` NACKs every 5 min | Probes (`ZPP >= 100`) start receive-only and only poll when no reading arrived within `SCAN_INTERVAL`. Push readings (`*#4*ZPP*0*T*3##`) keep flowing; a silent probe still recovers. Zone sensors unchanged. |
| #309 blocking `importlib.metadata` in the loop | The OWNd version is resolved once in the executor during `async_setup` and cached in `hass.data`; the WebSocket handlers read the cached value. |
| #310 deprecated `via_device` | `_ensure_cen_device` probes the registry signature once and passes `via_device_id` on cores that accept it, `via_device` on older ones. Identifiers are unchanged → no duplicate devices. |
| (no issue) MH200N `Could not send message *#16*0##` on every boot | Startup discovery (`*#2*0##`/`*#4*0##`/`*#16*0##`) now consults `GatewayProfile.supports_who()`; subsystems the profile doesn't advertise are skipped. Same "poll what the gateway can't answer" class as #308. |
| (no issue) trace export | Frames now carry `direction` (the card read a non-existent `dir` field), `iso_time`, `dimension`, `is_ack`/`is_nack`; a stringified `"None"` firmware is reported as empty. |

#311 (F454 alarm: 9 entities, no arm/disarm) is a design question about WHO=5 zone/sensor discovery and needs a bus trace first — reply drafted separately, not addressed here.

### Review findings fixed

- **WebSocket gateway lookup** is `entry.runtime_data`-first with `hass.data` as legacy fallback; a `mac` filter returns that gateway or nothing (the old fallback could hand the card another physical bus).
- **`myhome/bus_monitor/send` and `/clear` require admin.** Any authenticated user (kiosk token) could previously inject `*5*…##` alarm frames onto the SCS bus.
- **`ConfigEntryAuthFailed`** on password rejection → HA starts reauth itself and shows "Reauthentication required" instead of "Failed to set up". Other test failures → `ConfigEntryNotReady` (retry) instead of `return False`.
- **`async_unload_platforms`** with the result honoured; `hass.data`/`runtime_data` stay intact when a platform refuses to unload.
- **Services registered once in `async_setup`** (quality-scale `action-setup`), never unregistered.
- **`entry.runtime_data` set before platforms are forwarded.**
- **Removed the runtime OWNd self-installer.** Reloading `OWNd.*` in-process invalidated the class objects already bound by platform modules (`isinstance` dispatch silently broke). `manifest.json` is the single pin source; HA installs it. Validator, smoke runner and packager updated accordingly.
- **Removed the copy of the card into the user's `/config/www`.** `add_extra_js_url` + static path is sufficient.
- **`quality_scale.yaml` is honest now**: `runtime-data: in_progress`, `entity-translations: todo`, `repair-issues: todo` (repairs.py helpers are never called from production code).
- Dead `CONNECTION_CLASS`, pyproject URLs → OpenWebNet-HA, `.gitignore` for local artefacts, `pdf_output.txt` untracked.

### Test changes

- `tests/conftest.py`: `attach_gateway` fixture that registers a gateway the way `async_setup_entry` does (config entry + `runtime_data` + legacy mapping).
- `tests/test_issue_304_310_batch.py`: regression tests for #304, #307, #308, #310.
- `test_init.py`: parametrised reauth test (`password_error` / `password_required`) asserting HA opened the reauth flow; retry path; unload-failure path; cached version resolver. Self-installer / www-copy tests removed with the code.
- `test_websocket.py`: runtime_data-first, legacy fallback, no-substitution multi-gateway, behavioural admin test (`Unauthorized` for non-admin on send/clear, read-only commands still allowed).

### Verification

```
pytest tests --cov=custom_components.myhome      1335 passed, 100.0 %
python scripts/verify_ha_standards.py            SUCCESS
ruff check custom_components scripts tests       All checks passed
```

Local dev environment is HA 2024.3.3 / OWNd 2.0.0b6; CI runs the latest HA. Protocol assumptions were double-checked against the installed OWNd source (`get_next` None contract, WHO=4 dimension 0 vs 15 parsing, dimension 12/14 lighting events, `GatewayProfile.supported_who`).

**Live test — HA OS 18.2, Core 2026.9.2, Raspberry Pi 5, BTicino MH200N (auto-detected via WHO=13):** branch zip built with `scripts/package_release.py` and installed into `custom_components/myhome`. After restart: one `custom integration myhome` loader line, no `blocking call` (#309), no `via_device` warning from myhome (#310, sonoff still emits its own), no `not a message: None` (#304), static route `200 text/javascript`, Lovelace resource updated to the new content hash, bus-monitor card streaming with local-time stamps (#305), trace export shows `ownd_version: 2.0.0b6` from the cached resolver. Only remaining `myhome` log line was the MH200N `*#16*0##` NACK, fixed in the last commit.

> **Support note discovered during the live test.** The test box initially loaded *no* custom integrations at all (every card "Not loaded", static route 404, zero log lines). Cause: a previous b12 zip had been unpacked into the **root** of `/config/custom_components/` (stray `__init__.py`, `manifest.json`, …) plus a `myhome_backup_20260909/` copy inside `custom_components/`. On 2026.9.x a stray `custom_components/__init__.py` makes HA's loader skip every custom integration silently; the backup folder registered a second `myhome` domain. Neither is caused by this PR, but it's the same failure class as the `myhome.backup` caution in the README — worth a line there (follow-up).

### Follow-ups (not in this PR)

1. #307 part 2 — a per-light lock (`myhome.yaml` `rgb:` / `color_temp:` as authoritative, disabling auto-promotion) and a way to ignore the gateway's default `*#1*W*12*511*127*255##` / `*#1*W*14*1##` values.
2. #311 — WHO=5 discovery creates a panel per WHERE (system + zones + sensors). Needs a bus trace to decide zone → panel vs zone → binary_sensor.
3. Platforms still read `hass.data[DOMAIN][mac]`; migrate to `MyHOMEConfigEntry = ConfigEntry[MyHOMEGatewayHandler]` and finish `runtime-data`.
4. `has_entity_name` / entity translations (blocked on the entity-ID migration RFC).
5. Upgrade the local `.venv` to a current HA so deprecation removals are caught before CI.


GreenGrassBlueOcean:
we are going to split this as @fedem95 is not ready

GreenGrassBlueOcean:
Superseded by a stack of per-issue PRs, in merge order: #337 (the #335 privacy commits into v2) → #338 (init lifecycle, #309) → #339 (websocket lookup + admin gate) → #340 (#304, #310) → #341 (#308) → #342 (#307) → #343 (#305) → #344 (docs, quality-scale CI) → #345 (gateway identification, profile-gated discovery) → #346 (#302 cover, the #318 commits).

The tip of the stack is this PR's tree plus the #334/#335 catch-up it was missing (which is why it showed as conflicting); nothing here was dropped. One test needed adjusting on the way (#345, `dcd88c6`). Branch left in place until #319 is rebased onto `split/09-cover-302`.

Issue #302, opened 2026-09-12
RFC: timed covers, what I measured on a MyHOMEServer1 (latencies, frame order, non-linear position)


Interstellar0verdrive:
Over the last days I calibrated 12 shutters (two sizes, 198 cm and 107 cm) on a MyHOMEServer1 with my fork ([Interstellar0verdrive/MyHOME-stability](https://github.com/Interstellar0verdrive/MyHOME-stability)). Some of what I found is generic and may matter for the v2 cover model, so I'm writing it down here before proposing any code.

### 1. Bus timing (from the integration's own frame log, constant across runs)

- The motor starts **0.57 s** after the direction frame is written; the actuator's `*2*<dir>*<where>##` status marks that instant.
- The motor stops **0.08 s** after the stop frame is written.
- After every direction command the gateway relays, within ~0.1 s: `*2*1000#0*<where>##`, a real stop status `*2*0*<where>##`, `*2*1000#<dir>*<where>##`, and only then the moving status. The stop status is not a translation, so a translation filter (#283) does not cover it.
- The command session is serial: 0.1 to 1.2 s per frame depending on bus load. Twelve covers commanded together mean the last frame leaves the socket 1.5 to 12 s after being queued.

### 2. Three bugs a timed model gets from this, all found with a tape measure

- Clock started at enqueue instead of at the write: covers commanded together stopped 5–14 cm short.
- Echo window measured from enqueue: with a long queue the early stop status ended the run, and the later moving status started a free run to the end stop.
- A stop timer that had already fired when the clock was re-based: the stop went out right after the motor started.

I fixed them with delivery callbacks from the send queue (timestamp taken at the write), an echo window that spans the queue wait, and timers that carry a generation. Each has a test replaying the real frame sequence.

### 3. Position is not linear

Speed is proportional to the roll radius. With a linear model a 198 cm shutter sent to 50 % from open stopped at 85 cm from the floor instead of 99. One coefficient per direction (roll radius ratio, 1.5–2.3 on my shutters) plus a calibration in centimetres (two runs, a tape measure, two service calls) brought every shutter within 1–2 cm from the end stops and within 3–4 cm between intermediate positions. Docs: [configuration](https://github.com/Interstellar0verdrive/MyHOME-stability/blob/master/docs/configuration.md), [calibration recipe](https://github.com/Interstellar0verdrive/MyHOME-stability/blob/master/docs/recipes.md).

### What I can offer, and its maturity

[...]

Issue #283, opened 2026-09-11
WHO=1 Command Translation (`WHAT=1000`) frames are forwarded to entities and can set light/switch state to unknown


xtimmy86x:
## Description

While testing WHO=1 traffic, I noticed that OpenWebNet **Command Translation** frames such as:

```text
*1*1000#1*53##
```

are correctly recognized by OWNd as translation messages, but they are still forwarded to Home Assistant entities before being ignored by the gateway logic.

This can cause an existing `light` or `switch` entity to temporarily lose its state and become `unknown`.

It may also allow a translation frame to participate in dynamic light discovery even though it does not represent the actual state of an actuator.

## Example frame sequence

A typical physical ON command can generate:

```text
*1*1000#1*53##
*1*1*53##
```

The first frame is a **Command Translation**:

```text
WHO = 1
WHAT = 1000
WHAT parameter = 1
WHERE = 53
```

The second frame is the actual actuator state:

```text
WHO = 1
WHAT = 1
WHERE = 53
```

Therefore:

```text
*1*1000#1*53##
```

should not itself be interpreted as the actual ON state of light 53.

Likewise:

```text
*1*1000#0*53##
```

is a translated OFF command, while:

```text
*1*0*53##
```

is the actual OFF state.

## Current OWNd behavior

OWNd already appears to handle this correctly.

When parsing:

```text
*1*1000#1*53##
```

the message is classified as:

```python
_family = "COMMAND_TRANSLATION"
```

and:

```python
message.is_translation is True
```

`OWNLightingEvent` also deliberately does not assign `_state` when `WHAT == 1000`:

```python
if self._what is not None and self._what != 1000:
    self._state = self._what
```

As a consequence, for a Command Translation frame:

```python
message.is_on is None
```

This seems correct because the frame does not contain confirmed actuator state.

## Problem in MyHOME integration

The issue seems to be in the integration message routing.

In `gateway.py`, every parsed `OWNMessage` is first sent through the global dispatcher:

```python
async_dispatcher_send(
    self.hass,
    f"myhome_message_{self.mac}",
    message,
)
```

Only afterwards the code checks:

```python
if not message.is_translation:
    ...
else:
    LOGGER.debug(
        "%s Ignoring translation message `%s`",
        self.log_id,
        message,
    )
```

So the translation is ignored by the later gateway event logic, but it has **already been forwarded to the platforms**.

### Light platform

`light.py` listens to the global dispatcher:

```python
@callback
[...]

PR #318, opened 2026-09-14
fix(cover): anchor timed covers on the real write and motor start (#302)

GreenGrassBlueOcean:
## Summary

Fixes the three timed-cover bugs from #302, using the MyHOMEServer1 measurements @Interstellar0verdrive reported there and on #283. Also a correctness fix for `set_cover_position` on **any** gateway that relays the stop status after a direction command — previously the shutter ran to the end stop.

### What the bus actually does (measured)

```
enqueue ─(queue wait, up to 1.6 s with 12 covers)─▶ frame written
   +0.10 s  gateway relays a REAL stop status   *2*0*<where>##      ← not a translation
   +0.15 s  translation                          *2*1000#<dir>*<where>##
   +0.55 s  motor starts, direction status       *2*<dir>*<where>##
stop:       motor stops 0.08 s after the stop frame is written
```

### What was wrong

1. Clock started at **enqueue**, so a busy queue made every run up to 1.6 s (5–14 cm) short.
2. The relayed stop hit the "stopped" branch: it **cancelled the pending auto-stop**, froze the estimate, and the real direction frame 0.55 s later was then treated as a keypad press with no stop scheduled → shutter ran to the end stop.
3. A stop timer from a previous run could fire into a new run.

### Fix

**Gateway** — `send()` / `send_status_request()` return an `asyncio.Future` resolved with the monotonic time at which the sender worker writes the frame (queue wait included). Undelivered futures are cancelled on shutdown. Every existing caller ignores the return value and is unaffected.

**Timed cover** —
- Each command opens an **echo window**: from enqueue until `write + 1.5 s` (`ECHO_WINDOW`). Inside it a relayed stop *before* our own direction frame is ignored; our own direction frame **re-anchors the clock to the motor start** (absorbs the 0.57 s latency with no gateway-specific constant). An opposite-direction frame, or anything after the window, is a genuine external command and handled as before.
- `set_position` waits for the write, then for the motor-start echo (bounded by the window; falls back to the write time on gateways that don't relay it), and times the run from that anchor. When the timer fires the run is re-anchored *at the target* and the stop's own write time freezes the estimate — target plus whatever the queue delay added, e.g. 49 % instead of 50 % after a 0.1 s delayed stop at 10 s travel.
- The stop command freezes the estimate at its **write**, not at enqueue (`0.08 s` stop latency is noise; the queue wait isn't).
- Generation counter: a superseded auto-stop is a no-op.
- Without delivery info (legacy/mock gateway objects) the window is bounded from enqueue, so nothing is mistaken for an echo indefinitely.

Deliberate trade-off: a keypad stop pressed in the ≤0.6 s between our write and the motor start is treated as an echo. Physically the motor hasn't started yet, so nothing is lost; after the motor-start echo the window closes and keypad stops are honoured immediately.

### Tests

`tests/test_issue_302.py` drives the exact measured sequence with a controllable clock and real delivery futures: relayed stop survives, timer anchors on the motor-start echo, stop freezes at write (+overshoot), opposite direction inside the window is external, stale generation is a no-op, no-echo gateways anchor on the write, cancelled delivery falls back, gateway future resolves at write and is cancelled on `close_listener`. One existing cover test updated for the new semantics (stop right after our own open is an echo; after the motor-start echo it's genuine).

```
pytest tests --cov=custom_components.myhome      1346 passed, 100.0 %
python scripts/verify_ha_standards.py            SUCCESS
ruff check custom_components scripts tests       All checks passed
```

Not yet live-tested on hardware — see "Live test" below.

### Live test plan (MH200N or MyHOMeServer1, timed cover with `travel_time` set)

1. Bus monitor filtered WHO=2; `cover.set_cover_position` to 50 % from fully open.
2. Expect TX direction frame, RX `*2*0*`, RX translation, RX direction; **no stop sent** until `travel_time/2` after the RX direction frame; then TX `*2*0*`.
3. Position attribute after the run ≈ 50 (49 with a small queue delay), state `open`, no `opening` stuck.
4. While it moves, press stop on the wall: state must go to stopped immediately (window already closed by the motor-start echo).

### Follow-ups

- Roll-radius / non-linear position calibration (Interstellar0verdrive offered a service + sequences) — separate PR once the fixture traces are shared.
- `WRITE_TIMEOUT` / `ECHO_WINDOW` are module constants; make them gateway-profile values in OWNd if MH200N/F454 measurements differ.


Interstellar0verdrive (changes requested):
Thanks for picking this up, and for turning the measurements into a model instead of a constant: the delivery future, the window spanning the whole queue wait and the re-anchor on the actuator's own moving status are the right shape, and `tests/test_issue_302.py` replaying the frame sequence on a controllable clock is a pleasure to read.

Five things I would like changed before merge.

1. **`gateway.py:858`: the timestamp is taken before the send, not after it.** `_resolve_written()` runs before `_command_session.send()`. OWNd's `send()` reconnects and re-sends on `ConnectionResetError`/`IncompleteReadError`, retries a NACK three times, and returns `None` on any other exception; and this loop closes the session after `COMMAND_SESSION_IDLE_TIMEOUT` (15 s, line 838). So the usual cover command, more than 15 s after the last frame, is stamped before a full reconnect and handshake. Could you reconnect explicitly first, keep the pre-write `t0`, and resolve the future only after a successful `send()`, cancelling it otherwise, so that a frame that never reached the bus cannot start a run?

2. **`cover.py:334` / `:349`: the echo window can stay open for ever.** `_echo_until = inf` is cleared only by a generation-matching, uncancelled `_on_written`, but `close_listener()` cancels queued futures (`gateway.py:905`) and the worker returns without draining on a refused connection (`gateway.py:823`). `_handle_echo` then swallows every stop frame for that cover (line 635): it sits at "closing" and stops obeying the wall switch. Please close the window on cancel, exception or timeout, and bound it from enqueue.

3. **`cover.py:397`: `except (TimeoutError, CancelledError): pass` swallows the task's own cancellation.** `async_will_remove_from_hass` cancels `_stop_task` without bumping `_run_generation`, so on unload mid-run `_auto_stop` survives and stops a removed entity. Please re-raise unless `written.cancelled()`.

4. **`WRITE_TIMEOUT = 10.0` (`cover.py:56`) is below the worst case in #302**: twelve covers put the last frame 1.5 to 12 s out, so `_await_motion_anchor` times out and anchors on "now". The docs say ~1.6 s; that figure is the per-frame wait, not the queue.

5. **`_on_written` never writes state** (`cover.py:347-362`): after a stop the freeze is silent, and not every gateway relays the stop status that would repair it.

Noted without blocking: no `start_delay` fallback where the moving status never arrives (the run is then about 0.57 s too long), and no stop priority in the queue, so cover twelve's stop waits behind eleven other frames.

Interstellar0verdrive (changes requested):
Closing 2, 3, 4, 5 and the start-delay fallback: all verified at `d843a2e`, and the tests really exercise the scenarios (the bounded window, the removal mid-run, the repaint at the write). Thank you, that was quick and thorough.

One point stays open, and it's 1, for a reason outside your diff:

- **The explicit reconnect never fires.** `gateway.py:868` guards on `is_connected`, but in OWNd `2.0.0b6` (the pin) `close()` only closes the socket; `_connected` is set only inside `connect()`. After the 15 s idle close the flag is still `True`, so the branch is skipped, `written_at` is stamped, and the reconnect + handshake happen afterwards inside `send()`: the case the first review described. `test_send_future_is_stamped_after_reconnect_and_cancelled_on_failure` passes because the mock flips `is_connected` after a frame, which the real session doesn't do. Either guard on the socket actually being open (`_stream_reader is None` after `close()`), or track "we closed it" in the worker itself, and have the test model `close()` the way OWNd does.
- **`connect()`'s result is ignored.** It never raises: it returns `None` after its retries, or a dict with `"Success": False` on a negotiation error, and the loop falls through to `send()`, which then runs a second full `connect()` cycle. With an unreachable gateway each queued frame can hold the single worker for about two minutes; with a wrong password it's two extra negotiations per frame, the lockout the startup path (`gateway.py:818-826`) deliberately avoids by terminating the worker. Please check the result and cancel the frame's future (or terminate, as at startup) instead of falling through.

Two notes, not blocking, so this is one round:

- `cover.py:693` has `hasattr(self.async_schedule_update_ha_state, "assert_called")` in production code: a test-double check. `_publish_state()` already has the right `hass`/`platform`/`entity_id` guards; the echo path could just call it.
- A second command issued while the first frame is still queued (open, then stop a second later on a busy bus) bumps `_run_generation`, so the first `_on_written` becomes a no-op and the stop's freeze finds no clock: the position stays at the old value even though the cover moved between the two writes. Before this PR that case was estimated from the service call. Minor, and maybe fine to leave for the stop-priority change you mentioned.

Interstellar0verdrive (approved):
Approving. Point 1 is solved the right way at `9054229`: the guard on the streams is what OWNd's `close()` actually drops, `connect()`'s result is handled in both the refused and the gave-up case, and the test now models the library as it is (flag still `True` after the idle close, streams gone, `send()` reopening on its own). The `hasattr` check is gone too. Thanks for the two rounds.

One note for later, not for this PR: `_session_is_open` reads OWNd's private `_stream_reader` / `_stream_writer`. Since OWNd is ours as well, a public "socket is open" property there would let the integration stop depending on a private name.

PR #319, opened 2026-09-14
feat(frontend): bus-monitor trace/sweep workflow, self-describing exports, node test suite

GreenGrassBlueOcean:
## Summary

The bus-monitor card showed the state of the buffer but hid the workflow: it was not obvious how to "get a trace", an exported file could not say whether it was a passive trace or a Sweep Bus snapshot, and it ignored the active filters. This PR makes the capture workflow explicit and the exports self-describing, and puts the card under test.

### Workflow
- **Start Trace** clears the buffer and records passively (toggle: **Stop Trace** freezes the buffer, REC badge while recording). **Sweep Bus** clears the buffer and queries every subsystem; it resumes a paused stream first — seen live: 90 replies were dropped while paused.
- Export / Copy labels and the file name follow the chosen kind (`Export Trace` / `Export Sweep`, `Copy Capture` / `Copy Sweep`); **Clear** returns to trace mode.
- An **ⓘ** button in the title row opens an in-card explanation of the two captures, the export naming and the transmit bar.
- Raw-frame transmission is disabled until "I understand the risk" is ticked; the guard is enforced in `_sendCustomFrame` as well as in the DOM. Start Trace and Sweep Bus are read-only and say so.

### Exports
- File name `myhome_<kind>_<model>_<filter>_<timestamp>.json`; the file exports **what the user sees** (active WHO / WHERE / direction filters applied) and starts with a `capture` block: kind, `sweep_at`, filters, window (first / last, frames, buffer frames, buffer depth, truncated).
- One `CARD_VERSION` constant instead of three stale literals.

### Layout
- Two-row header: single-line title with status badge and help button; a full-width toolbar grouped by intent — capture (Start Trace, Sweep Bus), output (Export, Copy), stream (Pause, Clear) — one 32 px control height. Below ~750 px the old single row wrapped the title to three lines.

### Tests
- `tests/frontend/harness.mjs` loads the card under `node --test` with a minimal DOM stub; `tests/frontend/bus_card.test.mjs` covers the modes, labels, buffer semantics, the paused-stream regression, export naming and capture block, the transmit guard and the markup. `tests/test_frontend_card.py` runs the suite from pytest (node is on GitHub's ubuntu runners; skipped locally only when node is absent and `CI` is unset).


GreenGrassBlueOcean:
Thanks both — all points addressed on this branch (three commits after the rebase onto #318 round 2, which is already in).

**@xtimmy86x — calibration trace scoped to the gateway** (397921b, card part in ee99703)
- Every recorded frame now carries `gateway_mac` (the gateway that recorded it); `get_last_calibration_trace(gateway_mac)` filters on it.
- `ws_cover_calibration_trace` resolves the gateway exactly like the other commands: `mac` given → that gateway only, unknown `mac` → `not_found`; `mac` omitted → the primary gateway (first configured entry), never a substitute. The reply carries `mac` so the caller knows what it was filtered on.
- The card sends the request through `_wsPayload`, so it carries the card's configured gateway like every other call. The exported file keeps naming the gateway by prefix only (no full MAC in exports, as per #335).
- Tests: backend two-gateway test on the buffer (A's frames only, any MAC spelling, `""` is not "all"), WS test with two attached gateways covering given / omitted / unknown `mac`, and a node test that calibrates on two gateways and checks the export for gateway B contains none of A's frames and no full MAC.

**@Interstellar0verdrive**
1. **60 s cutoff guard, in the backend** (57d10e7): `_calibration_run` refuses a run ending inside the 59–65 s window (`CALIBRATION_CUTOFF_MIN/MAX`) the moment it ends — `failed` event, nothing stored, the remaining runs are not driven, so your plant fails after one 61.5 s run instead of storing 61 s after three minutes. The error names the value and points at the stopwatch. 58 s / 66 s runs still store (long shutter, not the limit). Holds for the service from automations and the developer tools, not just the card.
2. **Gate moved and re-keyed**: `legacyModels` is gone from the card. The actuator decides — the backend guard is the authority — and the card only pre-selects the stopwatch when a cover already carries the ~61 s signature; a MyHOMEServer1 with ordinary times gets auto mode offered and the backend refuses if the actuator can't do it.
3. **Stopwatch anchored on the bus**: the backend now measures every run of a timed cover from the motion anchor #318 introduced (write time / direction echo) to the stop's write time or the actuator's stop status, and exposes `motion_started_at` and `last_run_seconds` / `last_run_direction` / `last_run_ended_at`. The card's clock runs from the click only until the entity reports `motion_started_at`, then re-anchors; Stop & Save waits for the backend's `last_run_seconds` and saves that — queue wait and click-to-stop-frame delay are out of the number. The click clock is the fallback only when the bus never confirms the run, and the banner says which was saved. Test: a run with 2 s queue wait, 0.55 s motor start and a stop frame written 1 s after the click saves 15.0 s, not 17.5.

On "runs end on a user press": that is what the stopwatch path now is (press → stop frame → measured on the bus). The service keeps the bus-stop model with the guard; if you'd still rather see the timed runs of `calibrate_cover` end on a press as in your guided flow, I'd take that as a follow-up.

On moving the calibration tools out of the card into their own PR (#270): I'm fine either way — the trace/export commits and the calibration commits are separable on this branch. Say the word and I'll split it.

Full suite 1447 passed, 100% coverage; node suite 35/35.

xtimmy86x:
Thanks for addressing the review. I checked the gateway-scoped trace changes and the regression coverage: my concern about the backend export mixing frames from different gateways is resolved.

I agree with splitting this PR. Please keep the bus-monitor workflow and trace/export improvements here, and move the calibration services and UI into a separate PR, with the management tools going into the dedicated panel as discussed in #270.

The backend cutoff guard and bus-based timing are useful improvements, but calibration deserves a focused review: detecting the known 60-second cutoff does not by itself establish that the measured duration matches the shutter’s physical travel.

Once the split is ready and CI is green, I’m happy to finalize my review of the monitoring/export changes.

Interstellar0verdrive:
Yes, please split it: trace/export here, calibration in its own PR on top of #346. I'll read the calibration one when it's up, against the three claims (backend guard, actuator as the authority, stopwatch measured on the bus), the same way as before. On `calibrate_cover` ending on a press: a follow-up is fine, no hurry.

xtimmy86x:
Thanks for splitting this PR. My previous concerns are resolved, and the monitoring/export changes look good to me.

I also ran the full test suite locally on Python 3.14.4: **1,413 tests passed, 100% Python line coverage, and all 5 snapshots passed**.

One non-blocking follow-up: `_clearBuffer()` should reset `_traceStartedAt` so exports after Clear do not retain the previous trace’s start time.

Approved — ready to merge from my side.

GreenGrassBlueOcean:
Thanks for the review and the follow-up. The `_traceStartedAt` reset in `_clearBuffer()` is fixed in #351 (with a regression test), based on the same branch as this PR.

PR #323, opened 2026-09-14
chore: test against the core users run (Python 3.14, core 2026.9.2) and fix what it exposed

GreenGrassBlueOcean:
## Summary

Rebuilt on #316 alone, per the agreed queue (#316 → #323 → the rest one at a time). Nothing from #317–#322 is needed for it any more.

Home Assistant 2026.3+ needs Python 3.14.2. The workflows still installed an unpinned `homeassistant` on Python 3.12, which resolves to **2025.1.4** — a core nobody runs. CI now installs `pytest-homeassistant-custom-component` on Python 3.14 (it pins the exact core release it was built for, 2026.9.2 today) and runs the whole suite on it; the upstream-compat matrix keeps beta / dev.

### What the current core exposed, fixed here on the base

- Entities wrote state before their platform had added them (`KeyError: 'integrations'` on 2026.9). One `_publish_state()` guard on the base entity (`hass`, `platform` and `entity_id` all set, `RuntimeError` on a removed entity logged at debug) replaces the ad-hoc `hass is not None or mock` checks in every platform.
- `device_registry.devices` as a mapping and `async_get_device` are deprecated — identifiers are per config entry since 2026.8. The orphan-pruning lookup is scoped to the entry (`_device_for_identifier`); tests use `async_get` / `async_entries_for_config_entry`.
- `LightEntity.color_temp` is gone from the core: a mired accessor stays next to the Kelvin one so the DALI dimension-14 tests still inspect the bus value.
- Climate must advertise `TURN_ON` / `TURN_OFF` explicitly (mandatory since 2025.1).
- Test gateways built with `Mock(spec=...)` had no `device_registry_id` (18 failures); `via_device` on a core without `via_device_id` is now covered explicitly instead of by accident of the installed core.

`pyproject`: `requires-python >= 3.13`, classifiers 3.13 / 3.14, `syrupy` unpinned. CONTRIBUTING describes the uv-based setup (current cores import `fcntl`, so Windows means WSL).

Out of scope, kept as on the base: the `customize.yaml` recovery and its tests (removed later in #321), `entry.runtime_data` (#321). Those PRs shrink accordingly once this merges.

## Test plan

- [x] `pytest tests` on Python 3.14 / core 2026.9.2: 1328 passed, coverage 100 %
- [x] `scripts/verify_ha_standards.py`, `ruff`
- [x] Fixture and golden suites also pass on core 2025.1 (the CI this PR replaces)
- [ ] CI on this PR — first run on a 2026 core on GitHub; read failures, don't revert the bump


GreenGrassBlueOcean:
@Interstellar0verdrive, @xtimmy86x  why can i merge this one without review????

PR #331, opened 2026-09-14
refactor: frame router replaces the myhome_update dispatcher signals

GreenGrassBlueOcean:
## Summary

Third and last structural piece after the platform skeleton (#329, #330). Frame delivery used to be an agreement by string: a platform sent `async_dispatcher_send(hass, f"myhome_update_{mac}_{who}_{where}", msg)`, and every entity connected to the same f-string in its `async_added_to_hass` — 60+ places, each with its own list of address spellings to cover (`21`, `0021`, `21#4#01`, `general`, `0`). This PR puts that in one object.

No user-visible change. Entity ids, names, states and services are untouched; the full suite (1 455 tests, snapshots, golden corpus) passes.

### `router.py`

One `FrameRouter` per gateway on `entry.runtime_data.router`:

- `subscribe(who, keys, handler) -> unsubscribe` — a handler under any number of keys;
- `publish(who, keys, message) -> reached` — every handler under any of the keys, **once** per frame; a handler that raises is logged and skipped so one broken entity cannot starve the others.

### Who subscribes, who publishes

- The platform skeleton subscribes every entity it creates, under the same keys it remembers the device by (`known_keys`), **the moment it is built** — before `async_add_entities`. A discovery burst (the MH200N answers `*#2*0##` with eleven frames in a second) can no longer lose frames between creation and addition, which the old sensor platform worked around with a direct-feed mode; that mode is gone.
- The skeleton publishes each frame under the keys derived from it (`route_keys`), and the unsubscribe is tied to the entity's removal (`async_on_remove`).
- Cross-platform delivery goes through the same router: the light platform publishes WHO 1 frames of configured switches; covers subscribe under `general` and the cover platform publishes `*2*x*0##` there; alarm panels subscribe under `0` for the central unit's broadcasts; the audio pseudo zones `10x`–`14x` publish to amplifier `x`.
- Entities no longer subscribe themselves: the dispatcher blocks in ten `async_added_to_hass` methods (light, switch, cover, climate, alarm, media player, dry contact, aux, motion, illuminance) are deleted. `myhome_message_{mac}` (gateway → platforms) and `myhome_new_device_{mac}` (→ buttons) stay as dispatcher signals — they are not addressed to an entity.

### Tests

- New `tests/test_router.py`; the skeleton tests observe routing through the router.
- Nine assertions adapted: `async_on_remove` call counts (entities now register only the availability listener), and tests that dispatched on the old signal names now publish on the router — which is what the platform does.
- `tests/test_legacy_platforms.py` sent its "events" on signal names that never matched anything (`myhome_update_{mac}_1`, no key). They now go through the real bus feed, so the three snapshots that recorded `unknown` / `off` / `temperature: None` now record the frames arriving (`on`, `on`, `22.0`).

### Quality scale

- `strict-typing`: `router.py` is strict-clean; the migrated modules drop the ratchet **506 → 503**, baseline lowered.
- Coverage 100 % (`router.py` 42/42, `discovery.py` 246/246).
- `common-modules` comment and `docs/configuration/runtime_behaviour.md` (route step) updated.

## Test plan

- [x] `pytest tests` on core 2026.9.2 / Python 3.14: 1 455 passed, 1 skipped (node), coverage 100 %
- [x] `ruff check`, `scripts/verify_ha_standards.py`, `scripts/typing_ratchet.py` (503 ≤ 503)
- [x] `scripts/quality_scale_report.py`: unchanged — Bronze still waits on `brands` only
- [x] CI on GitHub once the account is unblocked
- [ ] Live on the MH200N: restart, then a bus sweep — every light, cover and contact state must update from the sweep replies (the routing path is what changed)

### DIFF
--- a/custom_components/myhome/climate.py
+++ b/custom_components/myhome/climate.py
@@ -33,6 +33,12 @@ from OWNd.message import (
     MESSAGE_TYPE_MODE,
     MESSAGE_TYPE_MODE_TARGET,
     MESSAGE_TYPE_TARGET_TEMPERATURE,
+    LOCAL_CONTROL_NORMAL,
+    LOCAL_CONTROL_OFFSET,
+    LOCAL_CONTROL_OFF,
+    LOCAL_CONTROL_PROTECTION,
+    LOCAL_CONTROL_OVERRIDE,
+    LOCAL_CONTROL_UNKNOWN,
     OWNHeatingCommand,
     OWNHeatingEvent,
 )
@@ -434,6 +440,7 @@ class MyHOMEClimate(MyHOMEEntity, ClimateEntity):
         self._attr_current_humidity = None
         self._target_temperature = None
         self._local_offset = 0
+        self._knob_pos = "UNKNOWN"
         self._local_target_temperature = None
 
         self._attr_hvac_mode = None
@@ -444,6 +451,8 @@ class MyHOMEClimate(MyHOMEEntity, ClimateEntity):
         """Return device specific attributes."""
         attrs = {
             "local_offset": self._local_offset,
+            "local_target_temperature": self._local_target_temperature,
+            "knob_pos":self._knob_pos,
         }
         if self._fan:
             attrs["fan_mode"] = self._attr_fan_mode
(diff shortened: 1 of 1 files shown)


GreenGrassBlueOcean:
Superseded by #365: same commit rebased onto `v2-phase1-architecture` (its base `feat/cover-travel-calibration` landed there through #363). GitHub refuses to change the base of a PR that is part of a stack, hence the new number.

PR #335, opened 2026-09-14
fix: privacy at the source - exports and diagnostics carry no personal data, and tests cannot

GreenGrassBlueOcean:
## Summary

Third part of the #316 split: the diagnostics and card redaction, stacked on #316 because the `--check` mode lives in the same script as the anonymizer. Retarget to `v2-phase1-architecture` once #316 is in.

- **Card** — the "Report Issue / Copy Trace" bundle and the JSON export no longer print the gateway's LAN address / port (or serial device path) or the browser user agent; they name the transport (`Ethernet TCP` / `USB / Serial`) and the gateway model, firmware and OUI only.
- **Diagnostics** — `TO_REDACT` grows from password / pin / token to `host`, `mac`, `id`, `UDN`, `ssdp_location`, `friendly_name`, `file_path`; the entry id and the user's title become placeholders. Home Assistant's own envelope (installed integrations, `setup_times`) is added outside our code and still needs the script — README says so.
- **The tree refuses it** — `scripts/anonymize_plant_fixture.py --check` reports LAN addresses, non-synthetic MACs, secrets, local paths carrying a user name and real config-entry ids; `test_fixture_privacy.py` runs it over the whole `tests/` tree; an optional pre-commit hook runs the same check. Two leftovers it found are fixed (`tests/golden/README.md`, `CONTRIBUTING.md`).

- **Redaction scope, settled** (from the review of #316) — `TO_REDACT` applies to the config entry's `data` and `options` only; the gateway, profile, queue, platform and bus-monitor blocks never pass through it, so every frame's `where` / `who` / `what` is there for triage. Our payload never carried device or entity ids (`platforms` is counts), and nothing in the download refers back to a redacted value (`id` is the formatted MAC, one entry ↔ one gateway), so no anonymized reference is needed — pinned by `test_redaction_is_scoped_to_the_config_entry`. The one user-named option, a decoder slot's `media_player.*`, becomes `media_player.decoder_<n>` so the slot → source / gain mapping stays readable.
- **The scan covers all of `tests/`** — the privacy test scanned `tests/fixtures` only (review); the root is now the `tests/` directory, with a regression case showing a dump under `golden/` or `frontend/` is caught.

## Test plan

- [x] `test_fixture_privacy`, `test_diagnostics`, `test_trace_replay`, `test_component_media_player` on core 2026.9.2
- [x] CI


xtimmy86x:
Thanks for addressing the review. I verified at `8a49dd6` that the privacy check now scans the entire `tests/` directory, with regression coverage for sensitive samples outside `fixtures/`.

The diagnostics redaction scope is now explicit and tested, and the decoder placeholders preserve useful configuration details without exposing personal entity names. All three CI checks pass.

My previous concerns are resolved. Approved — ready to merge from my side.

Issue #333, opened 2026-09-14
Climate entity for Zone 1 incorrectly changes to heating when another zone is heating


nicolacavallo84:
Good evening,

Integration version: 2.0.0b12
File usage: myhome.yaml
Gateway: MyHomeServer1

I would like to report a bug affecting the climate entities, which was already present in version 0.9.3 and earlier versions.

When any zone changes to the heating state, the first zone (climate.salone, zone 1) defined in my YAML file is also set to the same state, even though it is not actually heating at that moment. This only occurs with zone 1.

For this reason, I previously switched to the Michnovka fork  ([HomeAssistant-MyHOME](https://github.com/michnovka/HomeAssistant-MyHOME)), where this issue did not occur.

I am currently using the MHS1 gateway, but the problem also occurred with the H4890 gateway.

Heating system configuration:
Central unit: 3550
Probe: HC4692
Actuator: F430R8

Here is my configuration:
```
  climate:
    central_unit:
      zone: "#0"
      name: Centrale termoregolazione
      heat: True
      cool: False
      standalone: False
      manufacturer: BTicino
      model: 3550
    climate_salone:
      zone: "1"
      name: Salone
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
    climate_bagno_piccolo:
      zone: "2"
      name: Bagno piccolo
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
    climate_bagno_grande:
      zone: "3"
      name: Bagno grande
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
    climate_cameretta_sinistra:
      zone: "4"
      name: Cameretta sinistra
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
    climate_cameretta_destra:
      zone: "5"
      name: Cameretta destra
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
    climate_camera_matrimoniale:
      zone: "6"
      name: Camera matrimoniale
      heat: True
      cool: False
      standalone: False
      manufacturer: bticino
      model: LN4691
```

Interstellar0verdrive:
[...]

PR #336, opened 2026-09-15
fix(climate): an actuator status is not a second zone (#333)

GreenGrassBlueOcean:
## Summary

Fixes #333: the first climate zone switched to *heating* whenever any other zone started heating — since 0.9.3, on MyHOMEServer1 and H4890 alike.

A zone's actuator status frame, `*#4*<zone>#<actuator>*20*<state>##` ("zone 2, actuator 1 is on"), had its `#<actuator>` part read as a **second calling zone**, so the frame was also delivered to zone `<actuator>`. Nearly every zone has exactly one actuator, numbered 1 — so every "some zone started heating" frame also put zone 1 into `HEATING`. Only zone 1, only on heating start: the report word for word.

## Fix

One condition in `climate.py`'s calling-zone routing: the WHERE parameter is a zone only on a WHERE=0 frame (`*#4*0#5*…`, the form OWNd itself resolves to zone 5). On `<zone>#<n>` it is the actuator and is not routed anywhere else.

## Tests

`tests/test_issue_333.py` — the reporter's layout (central unit `#0`, zones 1-3, one actuator each):

- zone 2's actuator 1 turning on heats zone 2 and nothing else (fails on the unfixed code with `HEATING == IDLE` on zone 1)
- every zone reporting actuator 1: zone 1 follows only its own frame
- `3#7` discovers no phantom "zone 7"
- `*#4*0#2*0*0215##` (WHERE=0 with the zone in the parameter) still reaches zone 2

Full suite green on the core the base's CI runs (1333 passed).

## Branch

Based on `fix/anonymize-plant-fixtures` (#316) rather than on `v2-phase1-architecture`, only because the base is red until #316 merges (no plant fixtures); nothing here depends on #316 and it retargets to the base with no conflict once #316 is in. A bus trace from the reporter, run through the anonymizer, would make a good replay fixture — the existing `issue_247_myhomeserver1` plant is the same installation but has no actuator frames, which is why no replay caught this.


Interstellar0verdrive:
Leaving the review to @anotherjulien, heating isn't my ground. One note from the fork, since the two diagnoses differ on one frame: my fix for the same symptom routes every WHO 4 frame whose WHERE is `0` to the central unit (`4-#0`), because OWNd rewrites `*#4*0#1*20*1##`, the central unit's actuator 1, to zone 1. This PR keeps reading the parameter of a WHERE=0 frame as a zone and fixes the `<zone>#<actuator>` case instead. Both produce "zone 1 heats when anyone heats", so both may be real on the reporter's plant, or only one of them. A short bus trace from @nicolacavallo84 while another zone starts heating (anonymised with the script from #316) would tell which frames actually arrive, and would make the replay fixture you mention.

GreenGrassBlueOcean:
@fedem95 ready for review.

The reporter confirmed the fix on their plant (MyHomeServer1 + 3550, one actuator per zone): on the branch Zone 1 now stays idle while Zone 2 heats - see https://github.com/OpenWebNet-HA/MyHOME/issues/333#issuecomment-5693138043.

The temporary `2.0.0b13.dev0` marker used to tell the branch apart from the b12 release during the field test has been reverted; the tree is back to the two fix/test commits plus the manifest at `2.0.0b12`. Their first "still broken" report was a partial install (branch `climate.py` on b12's `myhome_device.py`, which lacks `_publish_state`) - not the fix.

Issue #16, opened 2021-07-18
Mh201 sometimes loose connection and reconnect


marcob79:
Sometimes it happen that command are not executed through my MH201, at first attempt, but if I retry it works.
Checking HA logs, I see some reconnections take place.
Maybe something to be improved in the connection management?
Please consider that the communication between HA and my MH201 is without WiFi, only Ethernet.

Thanks

This error originated from a custom integration.

Logger: custom_components.myhome
Source: /usr/local/lib/python3.9/site-packages/OWNd/connection.py:444
Integration: MyHome (documentation, issues)
First occurred: July 17, 2021, 4:38:56 PM (28 occurrences)
Last logged: 7:38:55 PM

Connection interrupted, reconnecting...

anotherjulien:
This behavior is unfortunately more common for some gateway than others.
The error you see is already the remediation in place in case the gateway refuses the connection. It will try to reconnect multiple times with increasing time between tries.
At this point there is not much more I can do in the integration since the problem is on the gateway. :-( 
Some "older" models are more prone to this problem than others; but at the same time there is also a lot of behavior variation within the same model with different firmware. I also suspect there are multiple hardware revisions of the same model number throughout the years that also behave differently...

caiosweet:
Yes correct.  Until version 7, disconnection occurred exactly every hour.  it was not a problem, but since version 8, it is constantly connecting / disconnecting ... today see to do debugging👍😇

anotherjulien:
As tested with @caiosweet this bug should be solved with 0.8.1, setting the "concurrent command workers" to 1 by default

marcob79:
Hi, thank you for your continue work in improving this component. It seems better but maybe i need some more hours to further confirm.
I was experiencing 4 new errors in log after restart that i don't understand what they are referred to, considering that i have an MH201 with just cover actuators.
one related to each of the following categories (that i don't have): binary sensor, sensor, switch and light

`Logger: homeassistant.components.binary_sensor
Source: custom_components/myhome/binary_sensor.py:116
Integration: Binary sensor (documentation, issues)
First occurred: 6:51:22 AM (1 occurrences)
Last logged: 6:51:22 AM

Error while setting up myhome platform for binary_sensor
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 249, in _async_setup_platform
    await asyncio.shield(task)
[...]

PR #338, opened 2026-09-15
refactor(init): entry lifecycle on the HA quality scale; no runtime OWNd installer (#309)

GreenGrassBlueOcean:
## Summary

Entry lifecycle aligned with the HA quality scale, plus #309. Split from #317 (part 2, on top of #337).

- **`ConfigEntryAuthFailed`** on password rejection → HA starts reauth itself and shows "Reauthentication required" instead of "Failed to set up". Other test failures → `ConfigEntryNotReady` (retry) instead of `return False`.
- **`async_unload_platforms`** with the result honoured; `hass.data` / `runtime_data` stay intact when a platform refuses to unload.
- **Services registered once in `async_setup`** (quality-scale `action-setup`), never unregistered.
- **`entry.runtime_data` set before platforms are forwarded.**
- **#309** blocking `importlib.metadata` in the event loop: the OWNd version is resolved once in the executor during `async_setup` and cached in `hass.data`.
- **Removed the runtime OWNd self-installer.** Reloading `OWNd.*` in-process invalidated the class objects already bound by platform modules (`isinstance` dispatch silently broke). `manifest.json` is the single pin source; HA installs it. Validator, smoke runner and packager updated accordingly.
- **Removed the copy of the card into `/config/www`.** `add_extra_js_url` + static path is sufficient.
- `quality_scale.yaml` honest: `runtime-data: in_progress`, `entity-translations: todo`, `repair-issues: todo`. Dead `CONNECTION_CLASS`, pyproject URLs → OpenWebNet-HA, `.gitignore` for local artefacts, `pdf_output.txt` untracked.

## Tests

`test_init.py`: parametrised reauth test (`password_error` / `password_required`) asserting HA opened the reauth flow; retry path; unload-failure path; cached version resolver. Self-installer / www-copy tests removed with the code.

## Verification

`pytest tests` 1335 passed; ruff clean. Live-tested on HA OS 18.2 / Core 2026.9.2 / MH200N as part of #317: no `blocking call` warning (#309), one `custom integration myhome` loader line, static route `200 text/javascript`.


fedem95 (dismissed):
Tested on an MH201 gateway: startup completed correctly, lights and covers initialized and responded normally.  
This also verifies the lifecycle fix: event and command workers start before platform setup, and partial gateway state is cleaned up before Home Assistant retries a failed connection.

GreenGrassBlueOcean:
@xtimmy86x would you please take a look at this PR neither myself nor fedem can check them since we both edited this PR

xtimmy86x:
Thanks for this cleanup. I reviewed `1536398` and found one startup ordering issue I would like addressed before approval.

Starting the workers before forwarding platforms makes sense to avoid blocking setup on a full command queue. However, `listening_loop()` also sends the initial discovery queries immediately. A fast gateway can respond while Home Assistant is still loading the platforms, before their discovery listeners are registered.

I reproduced this in an isolated simulation using the PR’s setup ordering, listener and cover-platform code: a cover response received before platform registration creates no entity; the same response received afterward creates it successfully. On a fresh installation, discovery could therefore remain incomplete until another bus event or a manual sweep.

Please keep the workers running early, but defer the initial discovery sweep until the platforms are ready, or buffer and replay early responses. A regression test should cover a fast gateway response during delayed platform setup while preserving the bounded-queue fix.

The remaining lifecycle changes look reasonable. CI is green, including 1,369 tests with 100% coverage, but I have not rerun the full HA suite locally or tested this on hardware.

fedem95:
Thanks @xtimmy86x . The discovery sweep was still inside listening_loop(), so it went out as soon as the listener started.
I moved it to initial_discovery(), which is queued from a background task after async_forward_entry_setups returns. Workers still start before the platforms.
The regression test uses a fake gateway that answers *#2*0## right away while platform setup is slow. It also shrinks the command queue to 2 and queues 6 status requests during that setup, so it covers the bounded-queue fix too. Without the discovery change the cover is never created. With the workers started after the platforms, setup hangs and the test times out. Full suite passes locally.
Also tried it with debug logging the three sweep requests are queued right after the last platform finishes setting up, behind the platforms' own status requests, and lights and covers respond normally.

GreenGrassBlueOcean:
@xtimmy86x I went through `bff2ae4` against your finding and it holds up:

- `listening_loop()` no longer queues anything on connect; the sweep lives in `MyHOMEGatewayHandler.initial_discovery()` and is queued from `async_setup_entry` only after `async_forward_entry_setups` returns, i.e. after every platform has connected its `myhome_message_<mac>` listener. The workers still start before the platforms, so the bounded-queue fix from `1536398` is preserved.
- Queuing it from a background task rather than awaiting inline is the right call: the sending workers don't drain the queue until the event session is ready, so an inline `await` on a queue the platforms had already filled could block setup.
- `test_fast_discovery_reply_during_slow_platform_setup` covers both halves you asked for: queue of 2, six status requests queued during a slow forward (only completes because a worker is already draining), and a fake gateway answering `*#2*0##` immediately - asserting no sweep during platform setup and that `cover` `…-2-31` exists afterwards.

CI is green on the new head and the suite also passes for me locally on core 2026.9.2 / Py 3.14. Ready for another look when you have a moment.

GreenGrassBlueOcean:
@xtimmy86x since you had the isolated simulation that reproduced the lost cover reply, could you re-run it against `bff2ae4`? With the sweep now queued after `async_forward_entry_setups`, the early `*2*0*<where>##` reply should create the cover in your setup as well. If you have a gateway handy, a fresh-install startup on hardware would be the ideal confirmation - `debug` logging on `custom_components.myhome` shows the three sweep requests being queued right after the last platform finishes. Thanks!

GreenGrassBlueOcean:
@xtimmy86x here is a hardware test recipe for `bff2ae4`, same shape as the one that worked in #333 (self-verifying install, no busybox `tar` filter, explicit rollback).

### Install the branch (Terminal & SSH add-on)

```bash
cd /config/custom_components
cp -r myhome myhome_backup_pre338
curl -sL -o /tmp/myhome-338.tar.gz https://github.com/OpenWebNet-HA/MyHOME/archive/refs/heads/split/01-init-lifecycle.tar.gz
rm -rf /tmp/myhome-338 && mkdir /tmp/myhome-338 && tar -xzf /tmp/myhome-338.tar.gz -C /tmp/myhome-338
rm -rf myhome && cp -r /tmp/myhome-338/MyHOME-split-01-init-lifecycle/custom_components/myhome myhome
grep -c 'initial_discovery' myhome/gateway.py; grep -c 'initial_discovery' myhome/__init__.py
```

Only if the last line prints `1` and `1` (b12 / master print `0` and `0`):

```bash
ha core restart
```

(`myhome.yaml` lives in `/config`, so it is untouched.)

### What to check

Add this to `configuration.yaml` before the restart so the ordering is visible:

```yaml
logger:
  logs:
    custom_components.myhome: debug
```

Then, ideally on a plant with at least one cover, zone or amplifier **not** declared in `myhome.yaml` (a fresh install is the strongest case), look for:

1. `Message `*#2*0##` was successfully queued.`, `*#4*0##`, `*#16*0##` in the log **after** the platforms' own status requests and after the last `Setting up myhome.<platform>` line - not right after `Creating listening worker.` as on `1536398`.
2. The undeclared cover / zone / amplifier appears as an entity on the first start, without a manual `myhome.sweep_bus` and without waiting for a bus event.
3. Nothing regressed on the bounded-queue side: on a large plant, setup completes and the entry reaches *Loaded* without a `ConfigEntryNotReady` or a hang.

If step 2 fails, please attach the startup section of the debug log (or a Bus Monitor trace export) - the `queued` lines vs. the first `*2*0*<where>##` reply will show exactly where the ordering goes wrong.

### Roll back

```bash
cd /config/custom_components && rm -rf myhome && mv myhome_backup_pre338 myhome && ha core restart
```

Thanks!

xtimmy86x (approved):
Thanks @fedem95 and @GreenGrassBlueOcean. I reran the isolated simulation against 9f81af1: the fast gateway response now creates the cover correctly, even with delayed platform setup.

I also tested on my real F454 and MyHomeServer1 gateways and can confirm it works. The startup logs show the three discovery requests being queued after platform setup on both gateways.

This resolves my startup discovery concern. I’m happy to approve, with merging subject to CI passing on the latest head.

GreenGrassBlueOcean:
Landed on `v2-phase1-architecture` through the cumulative #361 (every commit of `split/01-init-lifecycle` is an ancestor of `v2-phase1-architecture`; 0 commits ahead). Closing without merging; the remaining stacked work (#358 light feature lock) is being re-based onto `v2-phase1-architecture` as its own PR.

PR #339, opened 2026-09-15
fix(websocket): resolve the gateway from entry.runtime_data; admin-gate bus injection

GreenGrassBlueOcean:
## Summary

Bus-monitor WebSocket: gateway lookup and permissions. Split from #317 (part 3, on top of #338).

- **Gateway lookup is `entry.runtime_data`-first** with `hass.data` as legacy fallback; a `mac` filter returns that gateway or nothing. The old fallback could hand the card another physical bus.
- **`myhome/bus_monitor/send` and `/clear` require admin.** Any authenticated user (kiosk token) could previously inject `*5*…##` alarm frames onto the SCS bus. Read-only commands are unchanged.

## Tests

- `tests/conftest.py`: `attach_gateway` fixture that registers a gateway the way `async_setup_entry` does (config entry + `runtime_data` + legacy mapping).
- `test_websocket.py`: runtime_data-first, legacy fallback, no-substitution multi-gateway, behavioural admin test (`Unauthorized` for non-admin on send/clear, read-only commands still allowed).

The permissions note for `docs/configuration/bus_monitor.md` is in the docs PR later in the stack.


fedem95 (dismissed):
Tested on MH201 after reboot: 94 MyHOME entities loaded correctly, lights and covers work normally. WebSocket tests and lint are green. Approved.

GreenGrassBlueOcean:
Landed on `v2-phase1-architecture` through the cumulative #361 (every commit of `split/02-websocket` is an ancestor of `v2-phase1-architecture`; 0 commits ahead). Closing without merging; the remaining stacked work (#358 light feature lock) is being re-based onto `v2-phase1-architecture` as its own PR.

Issue #307, opened 2026-09-12
DALI DT8 Light Temperature and RGB Support


lyubomirtraykov:
I wold like to report that changing the color of the RGB works, but there are some problems:
- If the light is RGBW, when `*#1*WHAT*12*H*V*S##` is received color piker is activated, but when `*#1*WHAT*14*MIREDS##` is received after the light has been promoted to RGB, the color piker is disabled and the light becomes tunable withe only. The reverse is the same.
- I would like also to  report that `*#1*WHERE*12*511*127*255##` and `*#1* WHERE*14*1##` stays like that if they have never been set. If you send a massage `*#1*WHAT*#12*353*74*80##` to light that dose not support RGB, the gateway will accept it  and will save HVS values. The same is true if send `*#1*WHAT#14*153##` to RGB that dose not support tunable withe. The biggest problem is that you can't set `*#1*WHERE*12*511*127*255##` and `*#1* WHERE*14*1##` back. There should be a manuel way to specify the light supported features.
 
### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.9.1
- **Integration Version:** 2.0.0b12
- **OWNd Protocol Engine:** 2.0.0b6
- **Browser / User Agent:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36
- **Timestamp:** 2026-09-12T08:49:53.600Z

**Active Gateway Configuration:**
- **Model:** F461 (BTicino S.p.A.)
- **Firmware:** Unknown
- **Connection:** Ethernet TCP (10.11.1.149:20000)
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 1
- **Connection Status:** Connected

**Buffer Telemetry:**
- **Total RX Frames:** 211
- **Total TX Frames:** 86
- **Total Captured:** 297
- **Buffer Depth:** 68 / 200
- **Gateway Queue Depth:** 0
- **Active Card Filter:** None (All frames)

<details><summary>OpenWebNet Bus Trace</summary>

```
[08:43:07.648] [RX] *#1*25#4#02*14*419##
[08:43:07.832] [TX] *#1*25#4#02*#14*169##
[08:43:08.209] [RX] *#1*25#4#02*14*169##
[...]

PR #340, opened 2026-09-15
fix(gateway): skip OWNd's None reconnect cycle (#304); via_device_id for CEN devices (#310)

GreenGrassBlueOcean:
## Summary

Two small `gateway.py` fixes. Split from #317 (part 4, on top of #339).

| Issue | Fix |
|---|---|
| #304 MH201 `Data received is not a message: None` | OWNd's `get_next()` returns `None` for the cycle in which it reconnects (documented contract). The listener now skips that cycle at DEBUG instead of warning and dispatching `None`. The accompanying `Event connection lost, reconnecting...` line is OWNd's own reconnect log for the gateway-side idle close. |
| #310 deprecated `via_device` | `_ensure_cen_device` probes the registry signature once (`lru_cache`) and passes `via_device_id` on cores that accept it, `via_device` on older ones. Identifiers are unchanged → no duplicate devices. |

## Tests

`tests/test_issue_304_310_batch.py`, first section: None cycle skipped without warning; `via_device_id` when supported, `via_device` fallback; the probe reflects the installed core and is cached. The #308 and #307 PRs append their sections to the same file.

Live-tested as part of #317: no `via_device` warning from myhome, no `not a message: None`.


fedem95 (dismissed):
I pushed the tested gateway follow-up fixes to this branch.
The command worker now always accounts for queued items and closes its command session during shutdown, including after an unexpected send failure. The fixed 15-second idle close was removed, and commands use the lost-ACK retry path.
Routine None values received while an event session reconnects are now logged at DEBUG level rather than as warnings. CEN/CEN+ registration also waits for the parent gateway device and uses via_device_id.
I tested the changes on an MH201: lights, covers and CEN+ bus events work correctly.

xtimmy86x:
Thanks for adding the missing tests. I reviewed `421846b` and verified that all CI workflows are green, with 1,382 tests passing and 100% coverage.

I have no further blocking concerns with #340 and am happy to approve it. The stack divergence still needs to be resolved before merging the stack; this approval applies specifically to #340.

My review is based on the code changes and CI logs, without additional hardware testing.

xtimmy86x (dismissed):

GreenGrassBlueOcean:
Landed on `v2-phase1-architecture` through the cumulative #361 (every commit of `split/03-gateway-304-310` is an ancestor of `v2-phase1-architecture`; 0 commits ahead). Closing without merging; the remaining stacked work (#358 light feature lock) is being re-based onto `v2-phase1-architecture` as its own PR.

PR #343, opened 2026-09-15
fix(frontend): render bus-monitor timestamps in the browser's local time (#305)

GreenGrassBlueOcean:
## Summary

Fixes #305: bus-monitor timestamps 2 h behind. Split from #317 (part 7, on top of #342).

The card sliced the UTC ISO string. Frames now render through the browser's local zone (`_formatFrameTime`), matching the logbook. Exports keep the UTC values.

Card-only change; the doc note in `bus_monitor.md` is in the docs PR next in the stack. Verified on HA OS 18.2 as part of #317: card streaming with local-time stamps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #305, opened 2026-09-12
timezone issue


gimprota76:
Hello,

First of all, thank you for the great work you are doing on this project. Your efforts are very much appreciated.

I would like to report a timezone issue on my system.

The timestamps shown in the panel (Bus Monitor and Alarm events) are consistently 2 hours behind the actual local time in Italy (Europe/Rome timezone). It seems that the application is not synchronizing correctly with the Italian timezone or daylight saving time settings.

For example, alarm events and bus messages are displayed with a timestamp that is 2 hours earlier than the real local time.

Environment:
- Device: F454
- Feature affected: Alarm events and Bus Monitor timestamps
- Timezone: Europe/Rome (Italy)

Please let me know if additional logs, screenshots, or tests are required. I am available and happy to assist with further testing.

Thank you again for your support and for the excellent work on this project.

Best regards,
Giovanni Improta

<img width="505" height="722" alt="Image" src="https://github.com/user-attachments/assets/3853bf3f-bc96-4aec-8525-9ec904c24695" />

wave68runner:
You are faster than me , I was writing an issue and it was supposed to be the second bug I wanted to write. 😅

So I have the same issue with a MH200N. 
But I think the problem comes from HA but not from the gateway . For sure a variable utcnow() in place of now() or something like that

gdluck:
```
This error originated from a custom integration.

Logger: custom_components.myhome
Source: custom_components/myhome/gateway.py:287
Integration: MyHOME (documentation, issues)
First occurred: 23:52:03 (1 occurrence)
Last logged: 23:52:03

[MyHomeServer1 gateway - 192.168.0.9] Malformed event frame '*#13**0*23*52*03*999##': Invalid gateway timezone: '999'
```

GreenGrassBlueOcean:
@wave68runner @gdluck As a temporary workaround until the next beta release is out, you can fix this crash by logging into your physical MH200N / F454 gateway (via the web interface or MyHOME Suite) and explicitly configuring the Timezone setting.

The error happens because the gateway returns `999` (which means unconfigured timezone), crashing the current parser. Setting a real timezone will return a valid offset and bypass the crash!

GreenGrassBlueOcean:
@wave68runner @gdluck I've opened PR #387 which implements a permanent fix following Home Assistant's Golden Standard! 

[...]

PR #344, opened 2026-09-15
docs+ci+fix: runtime behaviour notes, quality-scale audit, gateway identification, timed covers (#302)

### LINKED ISSUE #302: RFC: timed covers, what I measured on a MyHOMEServer1 (latencies, frame order, non-linear position)

GreenGrassBlueOcean:
## Summary

Top of the #317 split stack, on top of #343 (frontend). It started as part 8 (docs and quality-scale tooling) and now also carries parts 9 and 10: #345 (gateway identification) and #346 (timed covers) were merged **into** this branch rather than into the base, so their commits live here and nothing else is left above it.

### Docs and quality scale (part 8)

- **README**: installation instructions with the backup-safety caution (a `myhome.backup` folder or stray files at the root of `custom_components/` make HA's loader skip every custom integration silently on 2026.9.x) and v2.0.0b12 links; live quality-scale tier table.
- **`docs/configuration/runtime_behaviour.md`** (new): profile-gated startup discovery, silent reconnect cycles (#304), reauth flow, push-driven probes (#308), additive colour modes (#307), `via_device_id` links (#310), OWNd version handling (#309), timed-cover echo model (#302).
- **`bus_monitor.md`**: time stamps (#305) and permissions (admin-only send/clear). **`gateways.md`**: identification notes.
- **`.github/workflows/quality-scale.yml`** + `scripts/quality_scale_report.py`: audit `quality_scale.yaml` against the official scale and report the tier reached; badge and README table kept live. Tests in `tests/test_quality_scale_report.py`.

### Gateway identification (part 9, was #345)

- **MH200N `Could not send message *#16*0##` on every boot**: startup discovery (`*#2*0##` / `*#4*0##` / `*#16*0##`) consults `GatewayProfile.supports_who()`; subsystems the profile does not advertise are skipped.
- **WHO=13 device type 4 is the MH200, not the MH200N**: `const.py` keeps the official and observed tables apart, cited.
- **Gateways are identified by precedence**: ssdp, serial, manual (config flow), who13; a model picked in the options flow outranks WHO=13. `identification()` is exported in diagnostics, the WebSocket info payload and the bus-monitor trace. Two repair issues (`gateway_identity_mismatch`, `gateway_identity_corrected`) with translations; stale identity warnings are cleared on reload.
- **Trace export**: frames carry `direction`, `iso_time`, `dimension`, `is_ack` / `is_nack`; a stringified `"None"` firmware is reported as empty.

### Timed covers (part 10, was #346, fixes #302)

- Timed covers are anchored on the real write and the motor start, with a bounded echo window; the write is stamped after the real send.
- The command session is reopened on the socket, not on `is_connected`.

### Packaging

- `check-wheel-contents` W002 is ignored in `pyproject.toml`. W002 flags byte-identical files; the identification change brought `translations/en.json` in line with `strings.json` (they had only differed by whitespace), which is the intended state for an HA integration.

## Tests

`tests/test_quality_scale_report.py`, `tests/test_gateway_identification.py`, `tests/test_issue_302.py` (all new); `test_gateway.py`, `test_repairs.py`, `test_trace_replay.py`, `test_component_cover.py`, `test_diagnostics.py`, `test_config_flow.py`, `test_discovery.py` (SSDP test kept off the network, TEST-NET-1 instead of real LAN addresses), and the discovery-gating section of `test_issue_304_310_batch.py`.

## Verification

`pytest tests` passes at this tip; ruff clean. Live-tested as part of #317: the MH200N `*#16*0##` NACK on boot is gone; gateway auto-detected via WHO=13.


GreenGrassBlueOcean:
Landed on `v2-phase1-architecture` through the cumulative #361 (every commit of `split/07-docs-quality-scale` is an ancestor of `v2-phase1-architecture`; 0 commits ahead). Closing without merging; the remaining stacked work (#358 light feature lock) is being re-based onto `v2-phase1-architecture` as its own PR.

Issue #292, opened 2026-09-11
Call to Action :: Gateway Traces :: #291


TheDarkWizard:
### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.9.1
- **Integration Version:** 2.0.0b11
- **OWNd Protocol Engine:** 2.0.0b6
- **Browser / User Agent:** Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36
- **Timestamp:** 2026-09-11T19:25:22.084Z

**Active Gateway Configuration:**
- **Model:** **_F454 (BTicino S.p.A.)_**  (This is not true btw, is a Myhomeserver1)
- **Firmware:** None
- **Connection:** Ethernet TCP (192.168.30.134:20000)
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 4
- **Connection Status:** Connected

**Buffer Telemetry:**
- **Total RX Frames:** 303
- **Total TX Frames:** 45
- **Total Captured:** 348
- **Buffer Depth:** 50 / 200
- **Gateway Queue Depth:** 0
- **Active Card Filter:** None (All frames)

<details><summary>OpenWebNet Bus Trace</summary>

```
[19:25:10.793] [RX] *1*1*12##
[19:25:10.794] [RX] *1*0*19##
[19:25:10.794] [RX] *1*1*12##
[19:25:10.794] [RX] *1*0*19##
[19:25:10.794] [RX] *1*0*13##
[19:25:10.794] [RX] *1*1*17##
[19:25:10.794] [RX] *1*0*13##
[19:25:10.794] [RX] *1*1*17##
[19:25:10.794] [RX] *1*1*14##
[19:25:10.794] [RX] *1*0*15##
[19:25:10.794] [RX] *1*1*14##
[19:25:10.794] [RX] *1*0*15##
[19:25:10.794] [RX] *1*0*16##
[19:25:10.795] [RX] *1*0*11##
[19:25:10.795] [RX] *1*0*16##
[19:25:10.795] [RX] *1*0*11##
[19:25:10.795] [RX] *1*1*07##
[19:25:10.795] [RX] *1*0*08##
[19:25:10.795] [RX] *1*1*07##
[19:25:10.795] [RX] *1*0*08##
[19:25:10.795] [RX] *1*0*09##
[...]

PR #345, opened 2026-09-15
fix(gateway): identify gateways by precedence; gate startup discovery on the profile

GreenGrassBlueOcean:
## Summary

Know which gateway you are on, then only poll what it supports. Split from #317 (part 9, on top of #344).

- **MH200N `Could not send message *#16*0##` on every boot** (no issue): startup discovery (`*#2*0##` / `*#4*0##` / `*#16*0##`) now consults `GatewayProfile.supports_who()`; subsystems the profile does not advertise are skipped. Same "poll what the gateway can't answer" class as #308.
- **WHO=13 device type 4 is the MH200, not the MH200N**: `const.py` keeps the official and observed tables apart, cited.
- **Gateways are identified by precedence**: ssdp (the gateway's own UPnP announcement), serial, manual (config flow), who13. Same family is consistent and variants are never downgraded. `identification()` is exported in diagnostics, the WebSocket info payload and the bus-monitor trace: model, source, raw code, what the spec and field evidence say it means, firmware / kernel / distribution, profile, conflict. Two repair issues (`gateway_identity_mismatch`, `gateway_identity_corrected`) with translations. The #292 replay test now uses the frame that gateway really sent (200) instead of a synthetic 2.
- **Trace export**: frames carry `direction` (the card read a non-existent `dir` field), `iso_time`, `dimension`, `is_ack` / `is_nack`; a stringified `"None"` firmware is reported as empty.

## Tests

`tests/test_gateway_identification.py` (new), `test_gateway.py`, `test_repairs.py`, `test_trace_replay.py`, and the discovery-gating section of `test_issue_304_310_batch.py`.

One commit is new relative to #317 (`dcd88c6`): #335's `test_redaction_is_scoped_to_the_config_entry` asserts the gateway diagnostics block verbatim, and this PR adds `identification` to that block. The test now gives the mock handler an `identification()` and asserts it arrives unredacted. #317 would have failed the same way once rebased.

Live-tested as part of #317: the MH200N `*#16*0##` NACK on boot is gone; gateway auto-detected via WHO=13.


xtimmy86x:
Thanks for this improvement. I reviewed `285eea2` and found two issues I would like addressed before approval.

**1. An explicit model selection does not reset `model_source`.**

The Options Flow updates `CONF_NAME` but preserves `model_source: who13`. The next WHO=13 response can therefore overwrite the user’s selection.

I reproduced this using the actual methods: an entry identified as `MH200` through WHO=13 is manually changed to `MH200N`; the next device-type `4` response changes it back to `MH200`.

Please record the manual selection as authoritative and add regression coverage for selecting the model, reloading the entry and receiving the next WHO=13 response. The selected variant should remain intact.

**2. A resolved identity warning can survive an integration reload.**

A new handler starts with `_identity_conflict = None`, while the previous warning can still exist in the issue registry. When identification finds no conflict, `_set_conflict(None, ...)` returns immediately because the in-memory value is unchanged, leaving the old warning behind.

Please ensure the no-conflict path removes any existing mismatch issue, even when the handler already holds `None`. A regression test should create the warning, recreate the handler and verify that a matching response clears it.

One small wording correction: the repair descriptions direct users to the reconfigure flow to change the model, but the model selector is in the Options Flow.

All three CI checks are green. These findings come from code inspection and isolated reproductions; I have not run the full Home Assistant suite or hardware tests.

GreenGrassBlueOcean:
Thanks for the careful review, both findings were real. Addressed in 96c8682:

**1. Explicit model selection is authoritative.** The options flow now writes `model_source: manual` whenever the model changes (`config_flow.py`), and `identification_source` honours it. Your repro (WHO=13-labelled `MH200` → user picks `MH200N` → next type-4 reply) now lands in the "manual, same family, leave alone" branch instead of relabelling.

**2. Stale warning across a reload.** `_set_conflict` still only *creates* the issue on a change, but the no-conflict path now always calls `async_delete_identity_issue` (a no-op when nothing is there), so a fresh handler holding `None` removes the previous instance's warning.

**3. Wording.** Both repair descriptions point to the integration's options (Configure) dialog now. The auth-failure text keeps "reconfigure flow", which is where the credentials actually live.

Regression coverage:
- `test_manual_selection_outranks_an_earlier_who13_label` — first reproduces the flip-back, then shows the manual-sourced entry stays `MH200N`.
- `test_stale_identity_issue_is_cleared_by_a_fresh_handler` — first handler raises the warning; a second handler on the same entry receives a matching type-4 reply and the delete fires.
- `test_options_flow_model_selection_survives_reload_and_next_who13` — end-to-end: real options flow on a `MockConfigEntry` with a pre-existing mismatch issue in the real issue registry → new `MyHOMEGatewayHandler` (the reload) → WHO=13 type 4 → `MH200N` kept, `model_source == manual`, stale issue gone.

Full suite on core 2026.9.2 / Python 3.14: 1415 passed.

xtimmy86x:
Thanks for splitting this PR. I reviewed: the remaining changes cover the bus-monitor workflow, trace/export improvements, documentation and frontend tests. Calibration is now outside this PR.

I ran the 12 frontend tests successfully. My previous concerns are resolved, and I approve the monitoring/export changes.

One non-blocking detail: `_clearBuffer()` should also reset `_traceStartedAt`, otherwise an export after Clear retains the previous trace’s start time.

Please ensure the full test/coverage workflow passes on this commit before merging.

PR #334, opened 2026-09-14
chore: test against the core users run (Python 3.14, core 2026.9.2) and fix what it exposed

GreenGrassBlueOcean:
## Summary

The content of #323, as its own PR against `v2-phase1-architecture` — split out of #316 at @Interstellar0verdrive's request so the toolchain and core-compatibility changes get their own reviewer. Same single commit, rebased on the base branch; nothing in it touches the fixtures.

- Workflows run on Python 3.14 / Home Assistant 2026.9.2 (the core users run) instead of the old pin.
- What that exposed, fixed: deprecated device-registry calls, state writes before the platform added the entity, the removed `LightEntity.color_temp`, and the test assumptions the eager-task scheduling in 2026.x invalidated.

## Test plan

- [x] `pytest tests` on core 2026.9.2 / Python 3.14 locally
- [ ] CI on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2026-09-15
fix(cover): anchor timed covers on the real write and motor start; echo window (#302)

GreenGrassBlueOcean:
## Summary

Fixes #302: timed covers. Exactly the commits reviewed and merged as #318 (both review rounds included), re-opened against the split stack. Part 10 of the #317 split, on top of #345. This is the top of the stack, so #319 rebases onto this branch.

- Timed covers are anchored on the real write and the motor start, with a bounded echo window; the write is stamped after the real send.
- The command session is reopened on the socket, not on `is_connected`.
- Echo model documented in `docs/configuration/runtime_behaviour.md`.

## Tests

`tests/test_issue_302.py` (new), `test_component_cover.py`.

## Verification

`pytest tests` 1412 passed at this tip; ruff clean. The tree here equals #317's tip plus the #334 / #335 catch-up that #317 was missing. Nothing from #317 was dropped (checked with `git diff` between the two tips).


Interstellar0verdrive (approved):
Approving: this is the same diff as #318 at `9054229`, both rounds included, compared line by line. Nothing to re-read.

PR #349, opened 2026-09-15
feat(cover): travel-time calibration on the bus - calibrate_cover service, buttons, bus-measured runs (backend; card UI in #355)

GreenGrassBlueOcean:
## Summary

The calibration half of #319, split out as asked there by @Interstellar0verdrive and @xtimmy86x — and, after their review here, split once more: **this PR is backend only** (services, buttons, cover model, WebSocket trace, tests, docs). The bus-card Covers panel and stopwatch are in #355 (draft, waits for the management panel #270); the card file and its node tests are untouched here. Commits are the reviewed ones from #319 with the card hunks removed (`a76b3f1` was card-only and is gone), plus two doc commits from the review. Sits on #319 (merged) → `split/09-cover-302`.

- `myhome.calibrate_cover` measures travel times on the bus: three timed runs per direction, anchored on the motion anchor #346 introduced, stored into the cover's `travel_time_up/down`. Calibrate / stop / set / reset buttons per cover. Every run of a timed cover is measured on the bus and exposed as `motion_started_at` / `last_run_seconds` / `last_run_direction` / `last_run_ended_at`, so a shutter whose actuator cannot report the run can be timed from Developer Tools: stop it by hand at the end stop, read `last_run_seconds`, store it with `myhome.set_cover_travel_time` (the stopwatch UI in #355 does exactly that).
- Every calibration frame is recorded with its `gateway_mac`; `bus_monitor/cover_calibration_trace` filters on it (9ecf3ca — xtimmy86x's #319 finding, already confirmed resolved there).
- Docs: `docs/configuration/services.md` (service, buttons, the Developer-Tools path), `runtime_behaviour.md` (what is measured). Per Interstellar0verdrive's review: the guard catches the factory 60 s cutoff only; an installer-set run time longer than the travel (30 s on a 14 s shutter) is stored as travel and only a stopwatch can tell — said in both docs.

## Interstellar0verdrive's three points from #319, and where each is answered

1. **Backend guard against the actuator's 60 s cutoff** (931baf8) — `_calibration_run` refuses a run that ends inside the 59–65 s window (`CALIBRATION_CUTOFF_MIN/MAX`): `failed` event, nothing stored, remaining runs not driven. A 14 s shutter on a MyHOMEServer1 fails after one 61.5 s run instead of storing 61 s after three minutes. Holds for the service from automations and the developer tools, not only the card. 58 s / 66 s runs still store.
2. **The actuator is the authority, not a gateway-model list** — no gateway-model list anywhere in the backend; the guard decides. (The card-side `legacyModels` removal now lives in #355.)
3. **Stopwatch measured on the bus** (45897b7, a76b3f1) — every run of a timed cover is measured from the motion anchor to the stop's write time or the actuator's stop status; the cover exposes `motion_started_at`, `last_run_seconds`, `last_run_direction`, `last_run_ended_at`. The card's clock runs from the click only until `motion_started_at` arrives, then re-anchors; Stop & Save saves the backend's `last_run_seconds`. Test: 2 s queue wait + 0.55 s motor start + stop frame 1 s after the click saves 15.0 s, not 17.5.

## Not in this PR (follow-ups, as agreed on #319)

- `calibrate_cover`'s timed runs ending on a user press instead of the bus stop (Interstellar0verdrive: "a follow-up is fine, no hurry").
- The card UI: #355 (draft). It lands in the management panel (#270) when that exists, not in the next beta.

## Tests

`tests/test_cover_calibration.py` (new: the runs and their anchors, a 61.5 s run refused at once, 58 s / 66 s runs stored, two-gateway trace scoping on the buffer and over WebSocket with given / omitted / unknown `mac`), `tests/test_component_button.py`. The node card suite is unchanged here (13/13); its +22 card tests are in #355.

## Verification

At this tip: `tests/test_cover_calibration.py` + `test_component_button.py` + `test_websocket.py` 62 passed; node suite 13/13 (base state). The Python tree equals the reviewed one; only the two doc commits are new. #355's tree equals the pre-split tip for the card file and its tests (`git diff` empty).


xtimmy86x:
Thanks for separating the calibration work from #319. @Interstellar0verdrive, this is the focused PR you offered to review for the backend guard and bus-based timing.

On the UI placement, I still think calibration belongs in the dedicated management panel, with the bus card focused on monitoring and trace exports. The panel is already under development, so please update the description accordingly.

I'm happy for the backend review to proceed now. Before merging the calibration UI, let's coordinate its placement in the panel so we avoid introducing tools into the card that we already intend to move.

Interstellar0verdrive (commented):
Backend read, against the three points I raised on #319, checked in the code and in the tests rather than the whole diff:

1. The 60 s guard is real and early: a 61.5 s run raises at once, one frame sent, nothing stored, `failed` event. That's the case that mattered on my plant.
2. `legacyModels` is gone; the backend decides. The "61 s signature" pre-selection in the card is harmless.
3. `last_run_seconds` is measured motor start to stop write (the 15.0 vs 17.5 test), the card re-anchors on `motion_started_at` and saves the backend's number, click clock only as a fallback. Good.

Nothing blocking from me on the backend. One limit worth a line in the docs: the guard catches the factory 60 s cutoff only; an actuator whose run time the installer set (30 s, say) stores that time as travel, and only the stopwatch can tell. The press-based runs stay a follow-up, as agreed.

On placement I'm with xtimmy86x: the stopwatch UI belongs in the panel, so I'd merge the backend and hold the card part, or land it knowing it moves.

GreenGrassBlueOcean:
Could we also include/allow the card part for now? That way we can push a new beta release and validate/test the backend in the meantime while @xtimmy86x works on the dedicated side panel. Once that’s ready, we can easily migrate the functionality over from the card to the sidebar.

xtimmy86x:
I understand the benefit of making the calibration tools available for beta testing, but I would prefer to avoid adding them to the card when we already intend to move them into the panel.

Calibration is useful, but it is not essential for the next beta. Shipping a temporary UI could mean users learn where to find it, only to have it moved or removed one release later.

My preference would be to merge the backend services and tests once ready, document the remaining measurement limitations, and introduce the calibration UI directly in the panel. Contributors could test the backend through Developer Tools in the meantime. The existing bus-monitor card would remain available.

@Interstellar0verdrive @anotherjulien @fedem95, what do you think? Would you agree with keeping the backend moving while holding the new calibration UI for the panel?

GreenGrassBlueOcean:
@xtimmy86x the question to ask is now how fast can you have a working version of the sidebar including the cover calibration? Would you be able to commit in the next couple of days

xtimmy86x:
@GreenGrassBlueOcean, a first version of the panel with guided cover calibration is already implemented in my [[draft PR](https://github.com/xtimmy86x/MyHOME/pull/1)](https://github.com/xtimmy86x/MyHOME/pull/1), so it is available for review and testing now.

It includes separate opening/closing measurements, explicit save to persistent travel profiles, Stop/Cancel handling, and a native bus monitor. The existing standalone card remains supported; its removal is deferred to a separate PR.

The remaining work includes aligning the calibration implementations. My branch currently uses backend-owned sessions with manual endpoint confirmation, while #349 introduces automatic bus-based calibration. We should agree on how these share the backend APIs and stored settings before integrating them.

I can make the current implementation available immediately, but I would not promise a merge-ready version within two days before completing that alignment, CI, and physical-gateway/browser validation.

This is why I would prefer to introduce the calibration UI directly in the panel. @Interstellar0verdrive, could we review the two approaches together and agree on the common backend?

GreenGrassBlueOcean:
@Interstellar0verdrive thanks — doc line added (72b10ca): the guard catches the factory cutoff only; an installer-set run time longer than the travel is stored as is, and only a stopwatch can tell.

On placement: split again as you and @xtimmy86x suggested. This PR is now backend only (the card file is untouched; e807e99 reworks the docs around the Developer-Tools path via `last_run_seconds` + `myhome.set_cover_travel_time`). The Covers panel and stopwatch are #355, draft until the management panel (#270) exists.

Interstellar0verdrive (approved):
Approving this one as it stands now: backend only, the three points from #319 verified in code and tests, the doc line on the guard's limit in. Thanks for the second split.

@xtimmy86x, yes on both: backend now, calibration UI in the panel, and let's agree the common backend before either side builds more. Having seen your draft (`cover_calibration.py` with manual end-stop confirmation, `cover_profiles.py`), it's close to what runs on my fork, so I'd sketch the shared shape like this:

- one calibration session owned by the backend (start / stop / confirm end stop), one per gateway, the panel and the service both clients of it;
- every run measured on the bus, motor start to stop write, as #346 and this PR do; how a run *ends* is the only variable: the actuator's stop status when it reports one (advanced covers), a user press otherwise, and the 60 s guard as the safety net for the automatic mode;
- results stored once, as profiles with a reference travel and per-cover overrides (opening and closing times, slat time, one roll coefficient per direction), with the origin of each value recorded; `cover_travel_times` from this PR would fold into that rather than stay a second store.

The contract of my fork's panel is on master (`docs/panel-websocket-api.md`), and the guided flow is documented in `docs/guided-calibration.md`, if that helps as a starting point. I'd rather write the contract down in #270 first and read your draft's backend against it than review two implementations side by side; I can do the reading, not this week.

PR #351, opened 2026-09-15
fix(frontend): reset the trace start time when the bus-monitor buffer is cleared

GreenGrassBlueOcean:
## Summary

Follow-up to the review of #319 (https://github.com/OpenWebNet-HA/MyHOME/pull/319#issuecomment-5681462037): `_clearBuffer()` reset `_lastSweepAt` but not `_traceStartedAt`, so an export made after **Clear** (without a new Start Trace) still reported the previous trace's `capture.started_at`.

- `_clearBuffer()` now nulls `_traceStartedAt` together with the frames. Start Trace sets it again after clearing, so that path is unchanged.
- Regression test: start a trace, clear, export → `capture.started_at` is `null`.

## Test

`node --test tests/frontend/bus_card.test.mjs`: 13 pass (12 existing + the new regression).


xtimmy86x:
Thanks for the follow-up. I verified that `_clearBuffer()` now resets `_traceStartedAt`, and the regression test covers Start Trace → Clear → Export.

All CI checks pass, including test-coverage. This resolves the remaining non-blocking point from #319.

Approved — ready to merge from my side.

Issue #303, opened 2026-09-12
[BUG] Standalone Climate Zones not exposing FAN


lyubomirtraykov:
I would like to report that the auto discovery is not exposing the FAN support for standalone climate zone.

[home-assistant_2026-09-12T07-11-33.608Z.log](https://github.com/user-attachments/files/32139178/home-assistant_2026-09-12T07-11-33.608Z.log)

### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.9.1
- **Integration Version:** 2.0.0b10
- **OWNd Protocol Engine:** 2.0.0b5
- **Browser / User Agent:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36
- **Timestamp:** 2026-09-12T07:14:03.993Z

**Active Gateway Configuration:**
- **Model:** F461 (BTicino S.p.A.)
- **Firmware:** Unknown
- **Connection:** Ethernet TCP (10.11.1.149:20000)
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 1
- **Connection Status:** Connected

**Buffer Telemetry:**
- **Total RX Frames:** 53
- **Total TX Frames:** 6
- **Total Captured:** 59
- **Buffer Depth:** 50 / 200
- **Gateway Queue Depth:** 0
- **Active Card Filter:** None (All frames)

<details><summary>OpenWebNet Bus Trace</summary>

```
[07:09:09.733] [RX] *4*303*1##
[07:09:09.796] [RX] *#4*1*0*0234##
[07:09:12.140] [RX] *4*4002#1*0#3##
[07:09:12.204] [RX] *4*4002*1##
[07:09:12.508] [RX] *4*4002#6*0#3##
[07:09:12.605] [RX] *4*4002*6##
[07:09:14.116] [RX] *#4*1#2*#20*5##
[07:09:14.180] [RX] *#4*1#2*20*5##
[07:09:14.220] [RX] *#4*1#2*20*5##
[07:09:14.516] [RX] *#4*6#2*#20*5##
[07:09:14.556] [RX] *#4*6#2*20*5##
[07:09:14.995] [RX] *#4*6#3*#20*5##
[07:09:15.044] [RX] *#4*6#3*20*5##
[...]

Issue #11, opened 2021-03-30
F454 , non riesco ad inserire le credenziali 


xman00007:
Salve, con il mio F454 , non riesco ad inserire le credenziali di accesso al F454 dal addon .. vanno inserite tramite .yaml ? grazie

anotherjulien:
I don't speak Italian.

xman00007:
Hi, with my F454, I can't enter the access credentials to the F454 from the addon .. must they be entered via .yaml? thanks

bubez81:
Did you try to restart homeassistant one or two times?

theBottinels:
check if your f454 as open connection (no authentication required) for some Ip range. If so disabled it.

PR #352, opened 2026-09-15
fix(climate): enable fan mode dynamically for climate zones with fancoils (#303)

GreenGrassBlueOcean:
## Summary
Resolves #303: Climate zones with fancoils (such as standalone thermostats wired to fancoils or F461 gateways without a 3550 central unit) did not expose fan speed controls or fan mode in Home Assistant.

### Key Changes
1. **Dynamic Fan Feature Activation (`_enable_fan_mode`)**:
   - Kept default `fan` support as `False` (radiator and floor heating zones are not given an unwanted fan speed selector).
   - Dynamically activates `ClimateEntityFeature.FAN_MODE` and sets fan modes `["auto", "low", "medium", "high", "off"]` as soon as any fan status event is observed on the bus (Dimension 11 or Dimension 20 actuator values >= 5).
2. **Fan-off State Tracking**:
   - Added `"off"` mode to supported fan modes and `fan_mode_map` (`"off": 4`).
   - Handled fan-off events (Dimension 20 `*#4*<zone>#<actuator>*20*5##` where `fan_on=False`, and Dimension 11 speed 4 / `fan_on=False`) so the entity transitions `fan_mode` to `"off"` instead of remaining stuck on the previous speed.
3. **Conservative HVAC Action Inference**:
   - Removed fallback setting of `HVACAction.HEATING` when valve is active.
   - When a zone is in `AUTO` or `OFF`, the action is left unchanged rather than falsely reporting heating in summer or when off. Only infers heating/cooling when the zone mode is explicitly `HEAT` or `COOL`.
4. **Actuator Isolation & Parameter Routing Fix**:
   - Actuator Dimension 20 fan status frames update fan speed and mode without corrupting valve `hvac_action`.
   - Prevent actuator index in WHERE parameters (`*#4*<zone>#<actuator>*...`) from being misrouted to other zones.
5. **Direct Fan Mode Service**:
   - Implemented `async_set_fan_mode` for climate zones, sending valid OpenWebNet frames (`*#4*<where>*#11*<speed>##`).
6. **Non-Mocked Real-World Trace Replay & 100% Statement Coverage**:
   - Added synthetic plant fixture `issue_303_f461` containing the anonymized 50-frame on-wire bus trace from issue #303.
   - Added deterministic non-mocked trace replay test in `tests/test_trace_replay.py`.
   - Added unit regression test suite in `tests/test_issue_303.py` covering all points raised in review.
   - Achieved 100.0% statement coverage on `custom_components/myhome/climate.py` (425/425 statements, 0 missing).
   - Fully sanitized and verified with `tests/test_fixture_privacy.py`. Zero sensitive data.


fedem95 (changes requested):
The dynamic part looks good to me: turning fan mode on when a fancoil frame shows up fixes [#303](https://github.com/fedem95/OWNd/issues/303) on its own, since the reporter's trace already has *20*5 / *20*8 frames.
The default is the part I'd change. is_standalone is already true for every zone that isn't #0, so fan = is_standalone and not is_central turns fan on for basically every zone. I checked with a plant that has a #0 central unit plus zone 1: zone 1 gets FAN_MODE on this branch and doesn't on v2-phase1-architecture. After the update everyone with radiators or floor heating would get a fan speed selector, and picking a speed sends *#4*<zone>*#11*<n>## to a zone with no fancoil. The central unit test doesn't catch this because it only checks #0 itself. I'd keep the default at False and rely on _enable_fan_mode().

A fan off frame (*#4*1#2*20*5##) sets fan_on=False with no speed, so fan_mode stays on the last value. After *20*8 then *20*5 it still says high.
With the valve active, the new fallback sets HEATING when the zone is in AUTO or even OFF. Before, the action was left alone. In summer on AUTO that shows heating. I'd only infer it from HEAT/COOL.

GreenGrassBlueOcean:
@fedem95 Thank you for the insightful review! All three points have been addressed in commit `56ae9c4`:

1. **Default fan support reverted to `False`**:
   - Standalone zones now keep `fan = cfg.get(CONF_FAN_SUPPORT, False)`. Radiator/floor heating zones will not expose a fan selector.
   - Zones with fancoils dynamically enable `ClimateEntityFeature.FAN_MODE` via `_enable_fan_mode()` when fan frames are observed on the bus.
   - Added test `test_plant_with_central_unit_and_zone_1_radiator_has_no_fan` reproducing your plant scenario (central unit + zone 1 radiator) to guarantee zone 1 does not expose fan support.

2. **Fan-off state tracking**:
   - Added `"off"` to supported fan modes (`["auto", "low", "medium", "high", "off"]`) and `fan_mode_map` (`"off": 4`).
   - Dimension 20 frames with `fan_on=False` (e.g. `*#4*1#2*20*5##`) and Dimension 11 frames with speed 4 or `fan_on=False` now transition `fan_mode` to `"off"` instead of remaining stuck on the previous speed.
   - Added test `test_fan_off_frame_after_high_transitions_to_off` verifying speed high followed by `*20*5` correctly transitions to `"off"`.

3. **Active valve HVAC action**:
   - Removed the fallback to `HEATING` when `MESSAGE_TYPE_ACTION` is received. In `AUTO` or `OFF` mode, `hvac_action` is left untouched rather than falsely reporting heating.
   - Added test `test_valve_active_in_auto_or_off_does_not_set_heating`.

All 1,348 tests pass with 100.0% statement coverage on `climate.py` (425/425 statements), and all GitHub Actions CI checks have completed green. Review re-request has been dispatched!

fedem95 (approved):
Thanks, that covers all three points. I re-ran the same checks I used to flag them, on 56ae9c4:
zone 1 in a plant with a #0 central unit no longer exposes FAN_MODE
*20*8 followed by *20*5 now moves fan_mode to off instead of staying on high
an active valve in AUTO or OFF leaves hvac_action untouched
Suite is green here too (1350 passed, 100% coverage), and your three tests assert what they claim, including the central-unit plant one.

PR #350, opened 2026-09-15
fix(button): sanitize entity_id with slugify for names with apostrophes (#347)

GreenGrassBlueOcean:
## Summary

Fixes #347: Lock and unlock button entities (`DisableCommandButtonEntity` and `EnableCommandButtonEntity`) and `MyHOMEEntity` were setting `self.entity_id` using naive string replacement `f"{platform.lower()}.{name.lower().replace(' ', '_')}_{action}"`. When a device name included apostrophes or accents (e.g. `"Chambre d'amis Groupe de volets"`, `"Chambre d'amis Volets arrière"`), this produced entity IDs with invalid characters (`button.chambre_d'amis_groupe_de_volets_unlock`), violating Home Assistant core entity ID format (`^[a-z0-9_]+$`) and triggering core validation warnings.

## Changes

1. **`custom_components/myhome/button.py`**:
   - Use `slugify` from `homeassistant.util` to sanitize device names into clean entity IDs (`button.<slug>_lock`, `button.<slug>_unlock`).
   - Graceful fallback to `slugify(f"device_{where}") or "device"` if device name is empty or results in an empty slug.
2. **`custom_components/myhome/myhome_device.py`**:
   - Apply identical `slugify` sanitization in `MyHOMEEntity.__init__` for `self.entity_id`.
3. **`tests/test_component_button.py`**:
   - Added `test_button_entity_id_sanitization_issue_347` covering names with apostrophes, accents, slashes, empty names, and dispatcher device discovery, asserting valid entity ID strings with `homeassistant.core.valid_entity_id`.

## Verification

- `pytest tests/test_component_button.py`: 7/7 passed.
- Full test suite: **1331 passed, 100.0% coverage** (5,275/5,275 statements).
- `ruff check custom_components tests`: clean (0 violations).
- `python scripts/verify_ha_standards.py`: all Home Assistant architectural standards passed.

PR #353, opened 2026-09-15
fix: restore actuator lock/unlock buttons after restart

xtimmy86x:
## Problem

Lock/Unlock buttons for automatically discovered lights, switches and covers can remain unavailable after a restart or integration reload. Home Assistant reports that these entities are no longer provided by the integration, while their parent entities continue working.

The parent platforms restore known actuators from the entity registry without emitting a new-device discovery signal. The button platform only creates buttons from configuration or that signal, leaving previously discovered buttons unloaded.

## Changes

* Rebuild Lock/Unlock buttons during setup from registered MyHOME lights, switches and covers belonging to the current config entry.
* Preserve canonical unique IDs, existing entity IDs, device associations and device metadata.
* Preserve leading zeros and secondary bus interface addresses.
* Deduplicate restored buttons against configured devices and subsequent discovery events.
* Avoid recreating buttons solely from orphaned button entries when their parent actuator is absent.

Existing buttons are recovered without requiring users to delete and recreate them.

## Validation

* **25 targeted tests passed**, including existing button and migration tests.
* New regression tests use Home Assistant’s entity platform to verify repeated setup, entity ID preservation, device metadata, gateway availability changes and secondary bus interfaces.
* Additional coverage verifies gateway isolation, WHO filtering and deduplication.
* Ruff and `git diff --check` passed.

Physical installation validation executed in my plant.


custom_components/myhome/button.py
```diff
@@ -98,6 +100,32 @@ def _create_buttons_for_device(dev_id, cfg):
     for _button in list(_configured_buttons.keys()):
         _buttons.extend(_create_buttons_for_device(_button, _configured_buttons[_button]))
 
+    # Discovered actuators are restored by their platforms from the registry.
+    # They no longer emit a new-device signal, so rebuild their buttons here too.
+    # Use parent actuators rather than stale button entries: a deleted actuator
+    # must not be resurrected just because its old buttons remain registered.
+    registry = er.async_get(hass)
+    for registered in er.async_entries_for_config_entry(registry, config_entry.entry_id):
```
GreenGrassBlueOcean: Non-blocking: platforms load concurrently, so this can run before `switch.py` prunes corrupted `…-1-1-06` unique_ids and before `light.py` removes "ghost" light entries whose WHERE belongs to a sensor/switch. In the `-1-1-` case the button ends up with `where="1-06"` and would send `*14*0*1-06##`. Only affects installs still carrying old corrupted registries, and the parent cleanup runs on the same startup, so low impact — just flagging it.

custom_components/myhome/button.py
```diff
@@ -98,6 +100,32 @@ def _create_buttons_for_device(dev_id, cfg):
...
+        who = {"light": "1", "switch": "1", "cover": "2"}.get(registered.domain)
+        if registered.platform != DOMAIN or who is None:
+            continue
+        prefix = f"{gateway.mac}-{who}-"
```
GreenGrassBlueOcean: Nit: `light.py`/`switch.py` strip either `gateway.mac` or `config_entry.data[CONF_MAC]` when parsing registry unique_ids; this only matches `gateway.mac`. They're the same string in practice, but accepting both would keep it consistent with the sibling platforms.

custom_components/myhome/button.py
```diff
@@ -98,6 +100,32 @@ def _create_buttons_for_device(dev_id, cfg):
...
+        if not registered.unique_id.startswith(prefix):
+            continue
+        device_id = registered.unique_id[len(prefix):]
+        where, _, interface = device_id.partition("#4#")
+        default_suffix = f"{where}I{interface}" if interface else where
+        device = dr.async_get(hass).async_get(registered.device_id) if registered.device_id else None
```
GreenGrassBlueOcean: Nit: `dr.async_get(hass)` could be hoisted out of the loop next to `er.async_get(hass)`.

tests/test_button_restore.py
```diff
@@ -0,0 +1,130 @@
...
+            await platform.async_reset()
+            await entry._async_process_on_unload(hass)
+
+
+async def test_restore_filters_and_deduplicates_actuators(hass):
+    """Keep gateway/WHO/address boundaries and ignore unrelated or deleted devices."""
+    entry = MockConfigEntry(domain=DOMAIN, data={"mac": MAC})
+    entry.add_to_hass(hass)
+    other_entry = MockConfigEntry(domain=DOMAIN, data={"mac": "00:03:50:00:00:02"})
+    other_entry.add_to_hass(hass)
```
GreenGrassBlueOcean, line 116: Nit: `entry._async_process_on_unload` is a private HA API — fine today, but the kind of thing that breaks on a core bump. Not a reason to hold the PR.

GreenGrassBlueOcean (dismissed):
Approving — small, well-scoped, and the diagnosis checks out against the code.

Verified locally:
- The registry-restore loops in `light.py` / `switch.py` / `cover.py` never send `myhome_new_device_*` (only the yaml and live-discovery paths do), so the premise is correct.
- The derived unique IDs / device identifiers match what `MyHOMEEntity` and the `__init__.py` migration produce, so restored buttons land on the existing device and keep their entity IDs.
- Merged onto current `v2-phase1-architecture` (66fcac0, post-#350 slugify change): clean merge, full suite passes in the 2026.9.2 / Py3.14 venv (1353 passed).

A few non-blocking notes inline.

xtimmy86x:
Thanks for the review and for running the full suite against the updated v2 branch.

I addressed the feedback in 99bc0c9:

* Malformed registry addresses such as `1-06` are rejected before buttons are created.
* Restoration checks sensor/switch ownership before processing lights, so it no longer relies on sibling platforms cleaning up ghost entries first. Matching preserves WHO, bus interface and leading zeros.
* Both gateway and config-entry MAC prefixes are accepted.
* Device-registry access is cached on first use.
* Tests now capture the public unload callbacks instead of calling `_async_process_on_unload`.

The new regression cases keep the corrupted entries present and exercise both registry orders, configured and registered sensors, and the commands generated by restored buttons.

All 30 targeted tests pass locally, along with Ruff and `git diff --check`.

GreenGrassBlueOcean:
super, accepted!

PR #355, opened 2026-09-15
feat(frontend): Covers panel and stopwatch in the bus-monitor card (card half of #349, waits for the management panel #270)

GreenGrassBlueOcean:
## Summary

The card half of #349, split out as asked there by @xtimmy86x and @Interstellar0verdrive so the backend can merge on its own. One commit, squashed from 1ed1a36 / 45897b7 / a76b3f1 on the original branch; the card code and its node tests are byte-identical to what was reviewed there. Sits on #349.

**Draft on purpose: this UI is agreed to belong in the management panel (#270).** Either this PR is reworked to land there once the panel exists, or it is closed and the panel PR takes the code. Not for the next beta — the backend is testable from Developer Tools (`myhome.calibrate_cover`, `myhome.set_cover_travel_time`, `last_run_seconds` on the cover) without it.

- 🪟 Covers panel: every timed cover with its up/down travel times, their source (`measured` / `manual` / `yaml` / `default`) and when they were measured; Calibrate per cover or all of them, with confirmation; live status; Export Trace (`bus_monitor/cover_calibration_trace`).
- ⏱️ Time → Stop & Save stopwatch for shutters whose actuator cannot report the run: the clock runs from the click only until `motion_started_at` arrives, then re-anchors on the bus; Stop & Save stores the backend's `last_run_seconds`, the click clock is only a fallback.
- The card pre-selects the stopwatch when a cover already carries the ~61 s signature; the backend guard stays the authority (no `legacyModels`).
- Docs: the Covers panel section in `bus_monitor.md`; the stopwatch mentioned next to the Developer-Tools path in `services.md` / `runtime_behaviour.md`.

## Tests

`tests/frontend/bus_card.test.mjs`: 35 pass (13 without this PR). No Python changes.


GreenGrassBlueOcean:
@xtimmy86x please take this as basis for the ux implementation.

xtimmy86x:
@GreenGrassBlueOcean, thanks for separating this work. I compared #355 (`970f267`) with my current panel implementation.

The panel already has guided calibration, reusable travel profiles, explicit review before saving, and backend-owned sessions. I would keep that workflow and use this PR as a reference for adding automatic calibration, measurement source/timestamps, and calibration trace export, once we agree on the shared backend.

I also reproduced four behaviors worth addressing before reusing the frontend code:

* `_timedCovers()` does not filter by the card’s selected gateway, so **Calibrate all** can include covers from other gateways.
* A single cover with a stored time in the 59–65 s range disables automatic mode for the entire list.
* Closing the stopwatch clears its timer without requesting Stop.
* If `stop_cover` fails and no new backend measurement arrives, **Stop & Save** can still persist the browser stopwatch value.

The existing 35 frontend tests pass; I checked these cases separately. For the panel integration, I would keep gateway scoping explicit, handle automatic-mode eligibility per cover, and require explicit handling of failed Stop requests and unconfirmed measurements.

The updated [[shared-contract proposal](https://github.com/xtimmy86x/MyHOME/blob/feat/myhome-sidepanel/docs/panel-shared-contract.md)](https://github.com/xtimmy86x/MyHOME/blob/feat/myhome-sidepanel/docs/panel-shared-contract.md) documents the remaining timing and persistence differences. We can use #270 to agree on those while keeping this PR as a reference for the additional UI functionality.

GreenGrassBlueOcean:
Superseded by #366: the same two commits on `v2-phase1-architecture` (base `feat/cover-travel-calibration` landed there through #363; GitHub refuses to change the base of a stacked PR). Still a draft for the same reason as before — waits for the management panel (#270).

Issue #370, opened 2026-09-16
[BUG] False gateway identity mismatch: confirmed F454 reports WHO=13 device type 200


xtimmy86x:
### Description

My gateway is a **BTicino F454**, confirmed from the physical device and correctly identified as F454 through SSDP. However, the integration logs an identity mismatch because WHO=13 device type `200` is mapped to `MyHomeServer1`.

### Warning

```text
2026-09-16 14:20:50.972 WARNING (MainThread) [custom_components.myhome] [F454 gateway - 192.168.100.70] Gateway identity mismatch: configured as F454 (ssdp) but WHO=13 device type 200 identifies MyHomeServer1 per field evidence.
```

### Relevant code

In `custom_components/myhome/const.py` on `v2-phase1-architecture`:

```python
WHO13_OBSERVED_DEVICE_TYPES = {
    "200": "MyHomeServer1",
}
```

The accompanying comment bases this mapping on field evidence from a MyHomeServer1 installation.

My F454 provides a counterexample: **observing code `200` on MyHomeServer1 does not establish that this code uniquely identifies that model.**

### Actual behavior

The integration reports an identity conflict even though the configured model matches both the physical device and its SSDP identification.

The SSDP model is retained by the current handler, but the warning incorrectly suggests that the gateway should be a MyHomeServer1.

### Expected behavior

* Keep the F454 identification supplied by SSDP.
* Do not report an identity mismatch solely because the device returns code `200`.
* Treat code `200` as ambiguous unless additional evidence reliably distinguishes the models.

### Suggested approach

Record the raw WHO=13 code for diagnostics, but avoid using `200` as a unique model identifier or as the sole basis for a mismatch repair.

This should also be considered before changing the corresponding mapping in OWNd: replacing a universal `200 → F454` mapping with a universal `200 → MyHomeServer1` mapping would still misidentify some devices.

Additional traces:

### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.10.0.dev0
- **Integration Version:** 2.0.0b12
- **OWNd Protocol Engine:** 2.0.0b6
- **Timestamp:** 2026-09-16T12:26:35.651Z

**Active Gateway Configuration:**
- **Model:** F454 (BTicino S.p.A.)
- **Firmware:** 2.0
- **Connection:** Ethernet TCP
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 1
[...]

Issue #13, opened 2021-05-06
How to manually configure a gateway


fcrisciani:
Hi,
In the doc is mentioned:
```
Some common gateways should be auto-discovered, but it is still possible to force the inclusion of a gateway not discovered. One limitation however is that the gateway needs to be in the same network as your Home-Assistant instance.
```
Can you provide an example of the field to put in the configuration.yaml?

anotherjulien:
Hi! I'll admit the doc is not very clear on that point...
Manually adding a gateway is not possible in the configuration.yaml file.

What can be done if the off-chance case where Home Assistant does not auto-discover the gateway is to go in the "Integrations" page and click "Add Integration", then select MyHome from there.
This will use the integration's own discovery mechanism that is very similar to the one used by Home Assistant, so it has the same limitations with regards to networking.

The reason for this is that going through this config flow is necessary for Home Assistant in order to create "Devices" and not just "Entities".

fcrisciani:
So unfortunately the extensions finds a gateway but it's not the one that I want to use and there seems to be no way to change the ip adress from the UI itself. Right now my current workaround is to change the config_flow.py and add the gateway manually in code. I was wondering if you can think to any better way to specify a gateway manually even if there is some auto discovered.

anotherjulien:
What's your gateway that is not discovered and what is the one that is?

fcrisciani:
So I have 2 myhome systems that are separate from the myhome bus point of view but they share the same network with the home assistant.
The gateway that is not discovered is an hold mh200 that is in the older system apartment.
```
<!-- G E N E R A L I N F O --> 
<date>30/07/2008</date> 
<webserver_type>MH200</webserver_type> 
<!--VERSIONE_FIRMWARE--> <ver_webserver>2.1.2</ver_webserver> <ver_webserver_hw>0</ver_webserver_hw> <!--ultima service pack installata--> <latest_sp>MH200_2_1_2.fwz</latest_sp>
<!-- G E N E R A L I N F O --> 
```

The other one is a new gateway f452 if I recall correctly the name

anotherjulien:
Can you show me what the discovery mechanism sees?
You need to run `python3.8 /usr/local/lib/python3.8/site-packages/OWNd/discovery.py` in the Home Assistant instance

fcrisciani:
This is the only things that finds:
```
bash-5.0# python3 /usr/local/lib/python3.8/site-packages/OWNd/discovery.py
[...]

PR #373, opened 2026-09-16
fix(gateway): treat WHO=13 device type 200 as ambiguous between F454 and MyHomeServer1

GreenGrassBlueOcean:
## Summary

Fixes #370 — F454 gateways reporting `*#13**15*200##` were incorrectly classified as MyHomeServer1, causing a false identity conflict warning on every connection.

## Root Cause

The WHO=13 dimension 15 device type table hard-coded code `200` → `"MyHomeServer1"`. In practice, **both F454 and MyHomeServer1 gateways** report code 200 (confirmed via community diagnostics and the reporter's own trace). When an F454 user saw code 200, the handler compared `gateway_model_family("MyHomeServer1")` against `gateway_model_family("F454")` → mismatch → false conflict.

### Core logic (`const.py`, `gateway.py`)
- Added `WHO13_AMBIGUOUS_DEVICE_TYPES` table mapping codes shared across gateway families (currently `{"200": ("F454", "MYHOMESERVER1")}`)
- Added `is_who13_code_compatible(raw_code, model_name)` helper that checks both unambiguous and ambiguous tables
- Rewrote `_handle_device_type()` to use compatibility checks instead of single-family string comparison
- Updated label for code 200 from `"MyHomeServer1"` to `"F454 / MyHomeServer1"`

### Documentation (`docs/configuration/gateways.md`)
- Updated WHO=13 dimension 15 table to reflect the ambiguity of code 200

### Tests & fixtures
- Added golden frame fixture `gateway.resp.model.device_type.200` for `*#13**15*200##`
- Added plant fixture `issue_370_f454/` with anonymized config and diagnostics
- Updated `test_gateway_identification.py`: compatibility tests, conflict resolution tests, table assertions
- Updated `test_trace_replay.py`: issue #292 test now expects no conflict when F454 receives code 200

## Verification

- ✅ **1603 passed**, 0 failed, 1 skipped (full test suite)
- ✅ Live MH200 gateway query confirmed code 4 = MH200 (not 200)
- ✅ OpenWebNet MCP validation of `*#13**15*200##` frame structure
- ✅ Ruff + HA standards checks clean


xtimmy86x:
Thanks for addressing #370. The change looks correct for my reported F454 case: code `200` is accepted without changing the configured model or profile, the previous mismatch repair is cleared, and an unknown gateway is no longer automatically labelled from this ambiguous code.

Before approval, I have two observations:

1. **Compatibility should distinguish “unknown” from “incompatible”.** The current list accepts F454 and MyHomeServer1, but still raises a mismatch for any other configured model reporting `200`. Our evidence confirms that these two models use this code; it does not establish that no other model can use it. Could we keep the configured identity and record the observation without a mismatch warning when compatibility is unknown? A confirmed contradiction could still trigger a warning.

2. **The CI failures are coverage-related.** In the runs I checked, the tests pass, but the strict coverage check fails on `const.py:271`, covering the observed, non-ambiguous code path in `is_who13_code_compatible()`. Please add a focused test for that path.

GreenGrassBlueOcean:
@xtimmy86x sorry I was too fast, previously this was not ready

PR #374, opened 2026-09-16
feat: add dedicated MyHOME administration panel

xtimmy86x:
## Purpose

MyHOME’s growing device-management and diagnostic features benefit from a dedicated interface. This PR introduces an experimental administration panel, keeping gateway setup in Home Assistant’s Config Flow and configuration validation and persistence in the backend.

This draft brings the work from [[xtimmy86x/MyHOME#1](https://github.com/xtimmy86x/MyHOME/pull/1)](https://github.com/xtimmy86x/MyHOME/pull/1) into the official repository for collaboration and review, following [[discussion #270](https://github.com/orgs/OpenWebNet-HA/discussions/270)](https://github.com/orgs/OpenWebNet-HA/discussions/270).

**This is work in progress, not a request for immediate merge or inclusion in the v2.0 release.** Development and fixes will continue on the source branch.

## Implemented

* Access through the integration’s Configure menu, with an optional sidebar entry.
* Gateway selection and inventory organised by WHO category, with category buttons and a selected-category/all-categories toggle.
* Entities grouped by device, collapsed by default, with A/PL addresses, search and filters.
* Native Home Assistant registry operations for names and areas.
* A bus monitor integrated into the panel, with trace capture, filtering, frame transmission and export. The existing standalone card remains supported.
* Cover travel profiles with separate opening and closing times, assignments, per-direction provenance and export.
* Guided, automatic and selected-cover batch calibration, with explicit review and save.
* Backend validation, revision checks and atomic profile persistence.
* Independent panel versioning, asset cache invalidation and English/Italian interface text.

## Next development step

The first shared backend task is to implement the agreed calibration schema, migration rules and a single resolver returning effective values and their origins.

Migration must preserve existing behaviour, including the relationship between native v2 timings and assigned panel profiles. Timing-only profiles remain valid without inventing a reference travel.

Subsequent work will add individual overrides and previews of shared-profile changes. Height scaling and nonlinear calibration will follow with their own runtime implementation and physical checks.

The covers section will retain the panel’s common WHO navigation, device grouping and interaction patterns. Dedicated calibration controls and an optional profile-grouped overview should extend that structure while keeping calibration directly accessible from a cover.

## Validation and review

The branch includes backend and frontend regression tests covering profile persistence, revision conflicts, gateway isolation, calibration lifecycle, inventory behaviour and panel loading.

The latest gateway compatibility fix also tests the installed OWNd command API directly, covering the incompatibility exposed after alignment with v2.

Physical verification of the combined v2/panel behaviour remains part of development. Automated tests do not establish calibration accuracy on real shutters.

Feedback is especially welcome on the backend contract, migration behaviour and consistency with Home Assistant’s native configuration model.


GreenGrassBlueOcean:
Really impressive work — thank you for bringing this into the official repo
and for keeping it as a draft while we figure out the shape together.

I have not gone through all 65 commits line by line; with a branch this
alive that would be the wrong kind of review anyway. The video is enough
to see that the panel is already a serious piece of product work. Gateway
overview, WHO inventory, A/PL, registry names/areas, cover profiles with
provenance, session-aware calibration, bus monitor in the same place —
that is exactly the management surface we sketched in #270. The backend
care (admin WebSockets, allowlisted inventory, Store revisions, atomic
writes, “don’t change travel times mid-motion”, don’t let native services
fight the panel) is what makes me comfortable discussing a preview at all.

What I would love us to keep aiming at is the most Home Assistant-native
version of this. Not because the current UI is wrong, but because the
integration will live next to Settings, device pages and Options Flow
for years, and those are the patterns users already trust.

A few things we could steer toward together:

- **Sidebar as an opt-in.** Configure → open the panel is the path that
  feels like HA. A sidebar shortcut is great for people who want it;
  default-off avoids surprising everyone else with another menu item.
- **The plant should keep working if someone never opens `/myhome`.**
  Discovery, names, areas, the existing cover buttons and the Lovelace
  bus card stay first-class. The panel then becomes the power tool, not
  a second control plane.
- **Lean on HA’s own UI where it already exists** — device page for
  names/areas when we can, `ha-dialog` for anything that moves a shutter
  or deletes a profile, and longer-term `hass.localize` so we are not
  maintaining a private en/it catalogue forever. No need to rewrite the
  frontend for that; just a direction.
- **Freeze the cover-profile schema for whatever we put in a zip.**
  v6 plus the current product rules is already a lot of thought. A
  preview is much easier to support if we are not also migrating storage
  mid-test-week.
- Small hardening when you have a moment: a public helper on the cover
  entity instead of private attributes, and a revision/etag on the
  inventory call so large plants are not shipping the whole catalogue
  every 15s.

None of that is “please start over.” It is “this is already good; let’s
make it feel like it was always part of Home Assistant.”

## Next beta (`2.0.0b13`)

I am lining up that protocol beta now (P7 resync, OWNd pin, repairs,
docs). Your PR text says this is not a request to include it in v2.0,
which I respect. I also do **not** want to leave you outside the test
window if you would rather have people try the panel on the same zip.

I am happy to include an **experimental, opt-in preview** in `b13` if
that is useful to you. Equally happy to leave it on this branch and take
it in the one after, if you would rather keep iterating.

If we did put it in this zip, I would only ask that we:

- ship a snapshot, not a moving HEAD
- keep the sidebar off unless the user turns it on
- freeze the profile schema for that tag
- call it experimental in the notes (calibration accuracy still needs
  real shutters; tests cannot sign that off — you already say this)

The protocol beta does not have to wait, and the panel is not being
sidelined either way.

@xtimmy86x @Interstellar0verdrive — what would help you more?

- include a frozen preview in `2.0.0b13`, or
- keep polishing on the draft and we pick it up in the next one?

Either answer is a good one. You have built something the project
actually needed; I just want us to land it in the Home Assistant way.

Interstellar0verdrive:
@GreenGrassBlueOcean — answering your b13 question.

On the panel in general, since you asked what would help. To be clear about what I'm commenting on: not the work, which is substantial, but the current blend of where the panel ends and where Home Assistant begins. That blend is what makes it feel early to me, more than any single screen does.

Concretely: some of what the panel offers already exists in native Home Assistant, and having both risks leaving the user unsure which is the real place ("where am I supposed to set the area?"). And in places Home Assistant's own dialogs open on top of the panel, which looks quite different, so the seam shows.

The rule I'd apply is this: the panel holds what is specific to BTicino and to the bus, and nothing about how Home Assistant configures and uses a device. It should give the extra, not a second way of doing what HA already does — we won't win on the second one, users already know HA's way.

Your own points about leaning on HA's own UI where it already exists, and about the plant working for someone who never opens `/myhome`, are the same concern from the other side. So my suggestion would be to define that separation first and ship the smallest possible version as an experimental proof of concept — only what this panel can do that Home Assistant can't — and grow from there.

@xtimmy86x — to be explicit, because none of the above reads properly without it: the work in there is genuinely impressive, and I'm not asking for a panel that looks like Home Assistant. A dedicated panel with its own custom interface is fine by me — for example Zigbee2MQTT and ESPHome both do exactly that, and any slightly advanced user has met it before. It's not my favourite pattern, but it's familiar and it works. What I think we should avoid is reinventing what Home Assistant already does well: that's where the separation of roles matters, and it would also be simply more to maintain — every device screen we build ourselves is one more thing that has to keep working as HA moves.

On covers specifically, which is the part I can speak about with some authority: the guided calibration is what I understood I was building, and I don't think what's there now is ready for anyone to see. At a quick look the parameters and the math all seem to be in place, which is the hard half done — but laid out like this it's a catalogue of parameters, and what has to sit on top of it is a flow: one thing at a time, in the order the shutter imposes, with the model staying behind the screen. Someone who opens this overlay for the first time closes it and doesn't come back, and we only get that first minute once.

So for b13 I would suggest: show covers the way we show lights and every other section, with the ability to see up and down travel times the v2 branch already has and the services that already set them, and leave the guided measurement flow out of this one.

One small thing that costs nothing now and a lot later: I'd avoid spending words like "profile" on what ships in b13. In the calibration those words carry a precise meaning, and a user who has already attached "profile" to something else has to unlearn it when the real one arrives. Anything that does survive in b13 is better named after what it does — the travel times — than after a concept we haven't shipped yet.

@xtimmy86x — two questions, just to check whether I've misunderstood something:

- Is the calibration flow still mine to build?
- Are you fine with leaving it out of b13?

And to say it plainly: nothing like this exists for MyHOME today, so the base is a real step up and the first sight of it is **impressive**. I'd just keep treating it as a proof of concept for now, rather than an almost finished package, and resist the temptation to show everything at once — experimental or not.

xtimmy86x:
> Really impressive work — thank you for bringing this into the official repo and for keeping it as a draft while we figure out the shape together.
> 
> I have not gone through all 65 commits line by line; with a branch this alive that would be the wrong kind of review anyway. The video is enough to see that the panel is already a serious piece of product work. Gateway overview, WHO inventory, A/PL, registry names/areas, cover profiles with provenance, session-aware calibration, bus monitor in the same place — that is exactly the management surface we sketched in #270. The backend care (admin WebSockets, allowlisted inventory, Store revisions, atomic writes, “don’t change travel times mid-motion”, don’t let native services fight the panel) is what makes me comfortable discussing a preview at all.
> 
> What I would love us to keep aiming at is the most Home Assistant-native version of this. Not because the current UI is wrong, but because the integration will live next to Settings, device pages and Options Flow for years, and those are the patterns users already trust.
> 
> A few things we could steer toward together:
> 
> * **Sidebar as an opt-in.** Configure → open the panel is the path that
>   feels like HA. A sidebar shortcut is great for people who want it;
>   default-off avoids surprising everyone else with another menu item.
> * **The plant should keep working if someone never opens `/myhome`.**
>   Discovery, names, areas, the existing cover buttons and the Lovelace
>   bus card stay first-class. The panel then becomes the power tool, not
>   a second control plane.
> * **Lean on HA’s own UI where it already exists** — device page for
>   names/areas when we can, `ha-dialog` for anything that moves a shutter
>   or deletes a profile, and longer-term `hass.localize` so we are not
>   maintaining a private en/it catalogue forever. No need to rewrite the
>   frontend for that; just a direction.
> * **Freeze the cover-profile schema for whatever we put in a zip.**
>   v6 plus the current product rules is already a lot of thought. A
>   preview is much easier to support if we are not also migrating storage
>   mid-test-week.
> * Small hardening when you have a moment: a public helper on the cover
>   entity instead of private attributes, and a revision/etag on the
>   inventory call so large plants are not shipping the whole catalogue
>   every 15s.
> 
> None of that is “please start over.” It is “this is already good; let’s make it feel like it was always part of Home Assistant.”
> 
> ## Next beta (`2.0.0b13`)
> I am lining up that protocol beta now (P7 resync, OWNd pin, repairs, docs). Your PR text says this is not a request to include it in v2.0, which I respect. I also do **not** want to leave you outside the test window if you would rather have people try the panel on the same zip.
> 
> I am happy to include an **experimental, opt-in preview** in `b13` if that is useful to you. Equally happy to leave it on this branch and take it in the one after, if you would rather keep iterating.
> 
> If we did put it in this zip, I would only ask that we:
> 
> * ship a snapshot, not a moving HEAD
> * keep the sidebar off unless the user turns it on
> * freeze the profile schema for that tag
> * call it experimental in the notes (calibration accuracy still needs
>   real shutters; tests cannot sign that off — you already say this)
> 
> The protocol beta does not have to wait, and the panel is not being sidelined either way.
> 
> @xtimmy86x @Interstellar0verdrive — what would help you more?
> 
> * include a frozen preview in `2.0.0b13`, or
> * keep polishing on the draft and we pick it up in the next one?
> 
> Either answer is a good one. You have built something the project actually needed; I just want us to land it in the Home Assistant way.

@GreenGrassBlueOcean Thanks for the thoughtful feedback. I would be happy to include a first experimental version in b13.

The many commits represent about a week of development, testing and incremental fixes. I appreciate that this makes the branch harder to review, and I agree that any preview should use an agreed, frozen commit.

On your individual points:

* **Configure and sidebar:** access through the integration’s Configure menu already exists, alongside the gateway settings. Sidebar visibility is already optional and persistent, but currently defaults to enabled. I agree with changing that default.
* **Working without the panel:** discovery, native entities, names/areas and the standalone bus card remain available without opening `/myhome`. The existing cover calibration buttons also remain. One qualification is that native timing writes are deliberately rejected while a panel profile/override owns those settings or a calibration session is active, to prevent conflicting changes. That interaction should be clearly documented.
* **Names and areas:** the panel already writes directly to Home Assistant’s native registries; there is no separate copy of that configuration. I find those contextual shortcuts useful when identifying devices by bus address, but I am open to discussing where linking to the native device page would work better.
* **Dialogs and translations:** confirmations are already present, but the dialogs use our own presentation around HTML `<dialog>`. I agree that adopting HA components where appropriate, and reviewing the localization approach, are useful improvements.
* **Profile schema:** agreed on freezing the format for the release. One update: the current storage schema is already v8, following the travel-scaling and nonlinear-model work, so that is the version we would need to review and freeze.
* **Public cover helper and inventory revisions:** both are valid points. There are still private-attribute accesses to clean up, and the inventory still returns the full catalogue on its periodic refresh.
* **Experimental status:** absolutely. The release notes should clearly distinguish implemented functionality from calibration accuracy verified on real shutters. Automated tests alone cannot establish that.

My preference would be to publish the current functionality as an explicitly experimental preview, after the agreed fixes and snapshot review. If a feature is implemented and behaves correctly, I would like us to identify a concrete reason before removing it from that preview.

If we feel the current scope needs more work before users should try it, I am equally comfortable waiting until it is more mature. I would prefer that to cutting working features simply to fit b13. The protocol beta can proceed independently.

xtimmy86x:
> @GreenGrassBlueOcean — answering your b13 question.
> 
> On the panel in general, since you asked what would help. To be clear about what I'm commenting on: not the work, which is substantial, but the current blend of where the panel ends and where Home Assistant begins. That blend is what makes it feel early to me, more than any single screen does.
> 
> Concretely: some of what the panel offers already exists in native Home Assistant, and having both risks leaving the user unsure which is the real place ("where am I supposed to set the area?"). And in places Home Assistant's own dialogs open on top of the panel, which looks quite different, so the seam shows.
> 
> The rule I'd apply is this: the panel holds what is specific to BTicino and to the bus, and nothing about how Home Assistant configures and uses a device. It should give the extra, not a second way of doing what HA already does — we won't win on the second one, users already know HA's way.
> 
> Your own points about leaning on HA's own UI where it already exists, and about the plant working for someone who never opens `/myhome`, are the same concern from the other side. So my suggestion would be to define that separation first and ship the smallest possible version as an experimental proof of concept — only what this panel can do that Home Assistant can't — and grow from there.
> 
> @xtimmy86x — to be explicit, because none of the above reads properly without it: the work in there is genuinely impressive, and I'm not asking for a panel that looks like Home Assistant. A dedicated panel with its own custom interface is fine by me — for example Zigbee2MQTT and ESPHome both do exactly that, and any slightly advanced user has met it before. It's not my favourite pattern, but it's familiar and it works. What I think we should avoid is reinventing what Home Assistant already does well: that's where the separation of roles matters, and it would also be simply more to maintain — every device screen we build ourselves is one more thing that has to keep working as HA moves.
> 
> On covers specifically, which is the part I can speak about with some authority: the guided calibration is what I understood I was building, and I don't think what's there now is ready for anyone to see. At a quick look the parameters and the math all seem to be in place, which is the hard half done — but laid out like this it's a catalogue of parameters, and what has to sit on top of it is a flow: one thing at a time, in the order the shutter imposes, with the model staying behind the screen. Someone who opens this overlay for the first time closes it and doesn't come back, and we only get that first minute once.
> 
> So for b13 I would suggest: show covers the way we show lights and every other section, with the ability to see up and down travel times the v2 branch already has and the services that already set them, and leave the guided measurement flow out of this one.
> 
> One small thing that costs nothing now and a lot later: I'd avoid spending words like "profile" on what ships in b13. In the calibration those words carry a precise meaning, and a user who has already attached "profile" to something else has to unlearn it when the real one arrives. Anything that does survive in b13 is better named after what it does — the travel times — than after a concept we haven't shipped yet.
> 
> @xtimmy86x — two questions, just to check whether I've misunderstood something:
> 
> * Is the calibration flow still mine to build?
> * Are you fine with leaving it out of b13?
> 
> And to say it plainly: nothing like this exists for MyHOME today, so the base is a real step up and the first sight of it is **impressive**. I'd just keep treating it as a proof of concept for now, rather than an almost finished package, and resist the temptation to show everything at once — experimental or not.

@Interstellar0verdrive Thanks for explaining your concerns. I agree that we should make the boundary with Home Assistant clear, and I would be happy to release a first experimental beta.

First, on your question about ownership: **yes, I still see the calibration UX as your part of our collaboration.** While implementing the backend, I needed a frontend to exercise the sessions, measurements, persistence and runtime behavior on my installation. That is why I built the current interface. It was not intended to replace your contribution or settle the final design. If the layout or interaction needs improving, that remains an area for you to lead within the shared panel structure.

However, I think we should distinguish the cover settings editor from the measurement flow. In the current version:

* Guided timing calibration proceeds through opening, closing, review and explicit save, showing the actions appropriate to the current step.
* “Slats and roll — guided” also provides instructions before each movement, intermediate stops, physical readings and repeat actions, with the calculations handled in the backend.
* Starting calibration replaces the settings form with that flow.
* Manual profile editing and advanced settings are initially collapsed.

So there is already a sequential guided flow, beyond the parameter editor. It may still need better wording, presentation or navigation, but I would find it helpful to walk through the current version together and identify the specific steps that need changing. I am not sure the description of a parameter catalogue captures what is currently implemented.

On the HA boundary, names and areas are updated directly in HA’s registries. I understand the concern about having two entry points and maintaining another editor, but I also see practical value in changing those details while identifying devices by their bus addresses.

Regarding “profiles”, the current implementation already has reusable profiles, assignments and shared-change previews. I agree that terminology must match the intended calibration model, but I would like us to clarify any actual mismatch before reducing that to “travel times”.

The commit count reflects about a week of development and testing; I am not presenting that as proof that the whole experience is finished.

On leaving calibration out of b13, my preference is to keep the implemented functionality in a clearly experimental preview, provided we agree it behaves correctly. If there is a specific correctness or usability problem that makes a feature unsuitable, let’s identify it and address it. I would not remove it solely because the interface can still improve.

If we are not comfortable publishing the current scope yet, I would rather continue refining it together and release the panel when it is more mature. There is no need to force it into b13.

Interstellar0verdrive:
@xtimmy86x — thank you, that's a clear answer and it's the one I hoped for. Building an interface to exercise the sessions on your own plant makes complete sense, and I read it that way now.

On b13 I'd take the second option you offered: keep refining it together and release the panel when it's more mature, rather than forcing it into this beta. The reason is the kind you asked for rather than a matter of polish — I couldn't get past the first step, and the log shows why. Not because it isn't worth it — this is the kind of thing that can genuinely change how people use MyHOME, which is exactly why I'd rather its first public appearance be a solid one.

**Slats and roll — guided.** Every time the session pushes an update, the panel throws while rendering and stops redrawing:

```
Error: Method not implemented.
  at scoped-custom-element-registry.ts:842
  at CoverCalibration._renderGeometry (panel-cover-calibration.js:217)
  at CoverCalibration._render (panel-cover-calibration.js:190)
  at CoverCalibration._accept (panel-cover-calibration.js:150)
```

31 times in the log, always the same line (`for (const input of form.elements)`). From the user's side it's a spinner that never ends. In my later attempts no opening or closing command reached the bus at all — only the stops I sent trying to get out of it.

The trace comes from the scoped custom element registry that Home Assistant's frontend loads, so it may depend on the HA version: frontend build `202608267` here, Chrome 153 on macOS. If you're on a different one you may simply never have hit it.

**Getting out of a stuck session.** Cancel closes unexpectedly the dialog, but still the session stays pending and is offered for resume on the next entry, and resuming lands back in the same state. I couldn't find a way to actually discard it.

**Guided timing measurement.** The first screen asks to close the cover before starting. I closed it with the wall switch and got "An unexpected movement interrupted the measurement" — before the measurement had started.

**First decision.** Three measurement modes are offered with no explanation of how they differ, and that's the first choice a new user has to make.

I stopped there. Happy to go through it again together once the render error is fixed.

On the parameter catalogue: you're right that starting a measurement replaces the form with the flow, and I shouldn't have blurred the two. But the form is still where the user arrives, and the measurement is one option among many values there — including slat time and opening and closing roll ratio, which can be typed in by hand. Nobody should be entering those manually: they only mean something as the output of a measurement. I'd keep them out of sight entirely and let calibrating be the one obvious thing to do on that screen.

One last thing, so my earlier suggestion isn't misread: the "stripped-down" preview was meant to spare all that effort from having to wait — outside covers, which is where I spent my time, the rest held up in what I tried. But if it's all or nothing, I'd rather wait for all of it.

xtimmy86x:
> @xtimmy86x — thank you, that's a clear answer and it's the one I hoped for. Building an interface to exercise the sessions on your own plant makes complete sense, and I read it that way now.
> 
> On b13 I'd take the second option you offered: keep refining it together and release the panel when it's more mature, rather than forcing it into this beta. The reason is the kind you asked for rather than a matter of polish — I couldn't get past the first step, and the log shows why. Not because it isn't worth it — this is the kind of thing that can genuinely change how people use MyHOME, which is exactly why I'd rather its first public appearance be a solid one.
> 
> **Slats and roll — guided.** Every time the session pushes an update, the panel throws while rendering and stops redrawing:
> 
> ```
> Error: Method not implemented.
>   at scoped-custom-element-registry.ts:842
>   at CoverCalibration._renderGeometry (panel-cover-calibration.js:217)
>   at CoverCalibration._render (panel-cover-calibration.js:190)
>   at CoverCalibration._accept (panel-cover-calibration.js:150)
> ```
> 
> 31 times in the log, always the same line (`for (const input of form.elements)`). From the user's side it's a spinner that never ends. In my later attempts no opening or closing command reached the bus at all — only the stops I sent trying to get out of it.
> 
> The trace comes from the scoped custom element registry that Home Assistant's frontend loads, so it may depend on the HA version: frontend build `202608267` here, Chrome 153 on macOS. If you're on a different one you may simply never have hit it.
> 
> **Getting out of a stuck session.** Cancel closes unexpectedly the dialog, but still the session stays pending and is offered for resume on the next entry, and resuming lands back in the same state. I couldn't find a way to actually discard it.
> 
> **Guided timing measurement.** The first screen asks to close the cover before starting. I closed it with the wall switch and got "An unexpected movement interrupted the measurement" — before the measurement had started.
> 
> **First decision.** Three measurement modes are offered with no explanation of how they differ, and that's the first choice a new user has to make.
> 
> I stopped there. Happy to go through it again together once the render error is fixed.
> 
> On the parameter catalogue: you're right that starting a measurement replaces the form with the flow, and I shouldn't have blurred the two. But the form is still where the user arrives, and the measurement is one option among many values there — including slat time and opening and closing roll ratio, which can be typed in by hand. Nobody should be entering those manually: they only mean something as the output of a measurement. I'd keep them out of sight entirely and let calibrating be the one obvious thing to do on that screen.
> 
> One last thing, so my earlier suggestion isn't misread: the "stripped-down" preview was meant to spare all that effort from having to wait — outside covers, which is where I spent my time, the rest held up in what I tried. But if it's all or nothing, I'd rather wait for all of it.

@Interstellar0verdrive Thanks for testing it and providing the error trace. The rendering failure is a concrete issue to investigate, and I appreciate you catching it.

I also want to clarify the readiness expectations for both you and @GreenGrassBlueOcean: **I have never said the panel was ready for release.** The PR is explicitly a draft and states that it is work in progress, with no request for immediate merge or inclusion in v2.0. I am not sure what gave the impression that I considered it nearly finished. My willingness to discuss an experimental preview was not a declaration of readiness.

In particular, I only implemented the **slats and roll guided flow yesterday**, and I had planned to verify it today. It is still fresh development, so your feedback is useful at exactly this stage.

On two of the reported points, I am attaching screenshots to clarify the intended behavior:

* **Preparing the cover:** the instructions explicitly say to start fully closed and stopped, and, if necessary, cancel and position it using the normal controls first. If the wall switch is used while the session is already active, the unexpected-movement interruption is therefore consistent with the current behavior. We can certainly make that preparation step more prominent or intuitive.

Screenshot 2026-09-18 144320


* **Cancelling:** there is a dedicated **Cancel measurement** button, separate from Stop, as shown in the second screenshot. I understand that you tried cancelling and were still offered the session again, so the button’s presence does not resolve your report. Could you confirm that this was the button you used and what exact session status appeared when you reopened the dialog? That should help distinguish a cancellation problem from a session waiting for Stop feedback.


Screenshot 2026-09-18 144347

I agree with continuing the work together and releasing the panel when we are comfortable with it, without tying it to b13. Thanks again for the testing and for clarifying that you did find the guided flow—the rendering issue is the first thing to address so you can actually complete it.

GreenGrassBlueOcean:
Thank you both for this thoughtful, constructive exchange. This is open-source collaboration at its best.

First off, @xtimmy86x — I saw you already tackled the `scoped-custom-element-registry` error in [`cb4f7d2`](https://github.com/xtimmy86x/MyHOME/commit/cb4f7d2260dbe7deecbba4dd343e39658ae35cc9) with direct query selectors. That was remarkably fast! And @Interstellar0verdrive, thank you for running it on your plant and providing the exact error trace and reproduction details — that kind of real-hardware testing is invaluable.

I want to make one thing crystal clear: **there is zero rush and zero pressure on either of you.**

We are shipping `2.0.0b13` independently this weekend with the core protocol updates (timezone 999 & unknown model repairs, WHERE 10 light availability, NACK logging, and the new documentation platform). There is already plenty for community testers to verify in `b13`, so we don't need the panel in this release at all. I will never ship or promote a major feature like this without both of you being completely comfortable with it.

Not publishing anything publicly right now also protects both of you from premature community bug reports while you are still actively shaping the design. As @Interstellar0verdrive rightly said, you only get that first impression once, and this feature is important enough that we should take the time to get it right.

So let's keep this strictly on the draft branch for now. You two take all the time you need to collaborate on the calibration flow and refine the interaction at your own pace.

Thank you both again for the fantastic work — enjoy your weekend!

Interstellar0verdrive:
@xtimmy86x @GreenGrassBlueOcean — thank you both, and that was a remarkably quick fix.

Fair point on readiness: that was my reading of the b13 question, not something you said.

**Preparing the cover.** Understood that the session is already active at that point. From the user's side, though, the screen had just asked me to close the cover, and doing exactly that ended the measurement — that's the part I'd design around.

**Cancelling.** Yes, it was Cancel measurement (video attached, recorded on the earlier build). Cancelling right away worked; if I waited instead, nothing happened for a while, then the dialog offered to resume a session I never interrupted, and from that point Cancel left it pending. My guess was that it came from the render error.

I've just done a quick check on the fix and the flow now gets going — the cover actually moves — so that was probably it. I can't test it properly in the next few hours, but I'll do a full pass as soon as I can and report back.

Video recorded on the earlier build:


https://github.com/user-attachments/assets/2aa5f593-9933-44ba-9841-0b76abeb48d7

xtimmy86x:
@Interstellar0verdrive Thanks for checking the fix. The cancellation behavior shown in your video was also related to the rendering bug you identified, which I fixed in [[cb4f7d2](https://github.com/xtimmy86x/MyHOME/commit/cb4f7d2260dbe7deecbba4dd343e39658ae35cc9)](https://github.com/xtimmy86x/MyHOME/commit/cb4f7d2260dbe7deecbba4dd343e39658ae35cc9). Thanks again for catching it and providing the details that helped track it down.

Regarding preparation of the cover, I’m completely open to your ideas and proposals for improving the user experience. The existing instructions describe the intended sequence, but we can certainly make the interaction clearer and more intuitive. Feel free to outline how you would approach that step—we can discuss it right here and work through it together.

xtimmy86x:
@GreenGrassBlueOcean Thanks for clarifying, that works well for me. Let b13 proceed independently, and we’ll keep developing and testing the panel on the draft branch.

I appreciate having the time to work through the calibration flow with @Interstellar0verdrive and verify it on our installations. His testing has already helped uncover a real issue, and that is exactly the kind of collaboration this draft was opened for.

We can revisit a public preview once we’re both comfortable with it. Thanks for the support, and enjoy your weekend too!

xtimmy86x:
@GreenGrassBlueOcean @Interstellar0verdrive In the meantime, I’ll continue improving and refining the panel to get it ready for a public preview, incorporating feedback and testing the changes on my installation.

I’ll keep the progress visible here so we can coordinate, particularly on the calibration UX, and build on each other’s work. Suggestions and proposals are always welcome.

Interstellar0verdrive:
@xtimmy86x following what we agreed on 16 September, here is the design preview for the calibration part of the panel: three paths (first of its kind / inherit a profile / fix a calibration), error states and the mobile layout. [I've put it in a Discussion](https://github.com/orgs/OpenWebNet-HA/discussions/419) so the wireframes have a stable place and this PR doesn't get buried.

It's a preview, not validated in code yet: I start implementing it in my fork on 27 September. Most of the open questions are about the panel itself (layout constraints for an overlay, whether you've already started anything on calibration, whether the session WebSocket contract of my fork can be the common base), so they're really for you. If you'd rather discuss it here, or in a call, that's fine with me too.

Issue #368, opened 2026-09-16
P7: lighting groups & general sync — declared groups + reactive re-sync for group/general frames


GreenGrassBlueOcean:
### Summary

Tracking issue for roadmap milestone **P7 — Lighting Groups & General Sync (WHO=1)**, scoped from the consultation in [discussion #248](https://github.com/OpenWebNet-HA/MyHOME/discussions/248#discussioncomment-18409209) and the DALI traces in #273 / #300. The design direction proposed there ([reply](https://github.com/OpenWebNet-HA/MyHOME/discussions/248#discussioncomment-18463857)): **Approach B (declared groups) scoped narrowly, with Approach C (reactive re-sync) as the fallback, and never an auto-discovered group entity.**

### Where the code stands today (`v2-phase1-architecture`)

| Frame | HA event | Status re-request | Entities updated |
|---|---|---|---|
| general `*1*x*0##` | `myhome_general_light_event` | none — `*#1*0##` is banned (rule in `.agents/rules/openwebnet-commands.md`, test `test_infeasible_commands.py`) | no |
| area `*1*x*A##` | `myhome_area_light_event` | `*#1*A##` after 0.1 s ([gateway.py:361-362](https://github.com/OpenWebNet-HA/MyHOME/blob/v2-phase1-architecture/custom_components/myhome/gateway.py#L361-L362)) | via replies |
| group `*1*x*#G##` | `myhome_group_light_event` | none | no |

Groups, areas and general are skipped by discovery ([light.py:327-329](https://github.com/OpenWebNet-HA/MyHOME/blob/v2-phase1-architecture/custom_components/myhome/light.py#L327-L329)); none of the DALI commits (#282, #285, #287, #289, #364) changed any of this.

### What the traces established

- **F461 + F429G echoes members.** A group dimension write `*#1*#1*#14*153##` is followed by `*#1*26#4#02*14*153##` and `*#1*25#4#02*14*153##` (#300). The light entities already consume those, so state is correct on this gateway without any membership map. A native group there buys single-frame broadcast only.
- **General dimension queries work on the F461.** `*#1*0*14##` / `*#1*0*12##` were answered by every ballast (#273). The current rule bans all WHO=1 general queries because `*#1*0##` NACKs on other gateways; it is coarser than the hardware and should be re-examined per gateway profile.
[...]

Issue #300, opened 2026-09-11
[FEATURE] Add support for lights groups


lyubomirtraykov:
Unfortunately F429G dose not support DALI groups. Every light has its own SCS address, and the only way to group them is by create SCS group of lights. When you send a request to the group address `#1` the getaway responds with the status of the group and every light in the group. I don't know if it is ok to create the group with the auto discovery or have a way to configure the groups manually. Most of the installations will group lights, so I think there should be  a way to configure and control groups. It is not specific for DALI only, the same behaviour is valid for normal and dimming light groups.  

### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.9.1
- **Integration Version:** 2.0.0b10
- **OWNd Protocol Engine:** 2.0.0b5
- **Browser / User Agent:** Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36
- **Timestamp:** 2026-09-11T21:38:23.156Z

**Active Gateway Configuration:**
- **Model:** F461 (BTicino S.p.A.)
- **Firmware:** Unknown
- **Connection:** Ethernet TCP (10.11.1.149:20000)
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 1
- **Connection Status:** Connected

**Buffer Telemetry:**
- **Total RX Frames:** 13
- **Total TX Frames:** 1
- **Total Captured:** 14
- **Buffer Depth:** 14 / 200
- **Gateway Queue Depth:** 0
- **Active Card Filter:** None (All frames)

<details><summary>OpenWebNet Bus Trace</summary>

```
[21:36:42.403] [TX] *#1*#1*#14*153##
[21:36:42.504] [RX] *#1*#1*#14*153##
[21:36:42.970] [RX] *#1*26#4#02*14*153##
[21:36:43.018] [RX] *#1*25#4#02*14*153##

```
</details>

GreenGrassBlueOcean:
Hi @lyubomirtraykov,

Thank you for the detailed feature request, system diagnostic bundle, and bus trace!

### 🔍 Architectural Analysis & Recommendation

For the time being, **managing lighting groups software-side using Home Assistant's native Light Group helper (`light.group`) is the officially recommended solution**. 

[...]

Issue #14, opened 2021-05-28
Unable to setup integration


monsterciattolo:
Hi,
i updated plugin to latest version 0.7.7 but i cannot setup again and i got this error.

Logger: homeassistant.config_entries
Source: custom_components/myhome/gateway.py:152
Integration: MyHome (documentation, issues)
First occurred: 14:07:26 (1 occurrences)
Last logged: 14:07:26

Error setting up entry MyHomeServer1 Gateway for myhome
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 269, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/config/custom_components/myhome/__init__.py", line 76, in async_setup_entry
    await myhome_gateway.connect()
  File "/config/custom_components/myhome/gateway.py", line 152, in connect
    await self.event_session.connect()
  File "/usr/local/lib/python3.8/site-packages/OWNd/connection.py", line 425, in connect
    await self._negotiate()
  File "/usr/local/lib/python3.8/site-packages/OWNd/connection.py", line 223, in _negotiate
    raw_response = await self._stream_reader.readuntil(OWNSession.SEPARATOR)
  File "/usr/local/lib/python3.8/asyncio/streams.py", line 629, in readuntil
    raise exceptions.IncompleteReadError(chunk, None)
asyncio.exceptions.IncompleteReadError: 0 bytes read on a total of undefined expected bytes

Thanks,
Marcello

anotherjulien:
Hi,

When did you update to 0.7.7?
Can you tell me which version of OWNd is required in your manifest.json file?

Thanks!

monsterciattolo:
Ownd version 0.7.26

anotherjulien:
Can you uninstall and reinstall version 0.7.7?
The OWNd dependency should be 0.7.27 (I removed version 0.7.26 because of a bug that is likely the cause of your problems)

monsterciattolo:
i followed your suggestion and now integration seems loaded but covers don't work

Logger: homeassistant.helpers.service
Source: helpers/service.py:128 
First occurred: 18:00:30 (7 occurrences) 
Last logged: 21:30:58

Unable to find referenced entities cover.letto
Unable to find referenced entities cover.soggiorno_grande

this is my configuration that worked before latest upgrade

- platform: myhome
  devices:
    bagno_beige:
      where: '14'
      name: Tapparella Bagno Beige
[...]

Issue #12, opened 2021-04-12
Myhome Log File


dirk291:
Julien,

Hereby the File.
Dirk
 
[Logfile.txt](https://github.com/anotherjulien/MyHOME/files/6297025/Logfile.txt)

PR #376, opened 2026-09-16
feat: implement declared groups as assumed state lights with member inference (#368)

GreenGrassBlueOcean:
## Summary

Adds **declared lighting groups** as a new kind of light entity: `where: '#G'` in
`myhome.yaml` declares a group that already exists in the plant (programmed with
MyHOME_Suite or a physical group-programmed actuator) and gives it a Home Assistant
light entity for single-frame control (`*1*1*#G##`, or `*#1*#G*#14*<mireds>##` for a
DALI colour-temperature change — the #300 case). Part of #368 (P7 — lighting groups
& general sync), design settled in discussion #248: **Approach B, scoped narrowly,
never an auto-discovered group entity.**

## End goal

- A group is **declared**, not discovered: OpenWebNet has no command to read back a
  group's membership, so the plant (not the bus) is the source of truth. `myhome.yaml`
  is the same path as `lock_features` (#364) for this kind of plant knowledge the bus
  can't provide.
- Without `members`, the entity is `assumed_state = True`: Home Assistant shows
  separate On/Off controls and the entity never claims to know the group's real
  state — only what was last sent to it.
- With `members` (each member's own point-to-point `WHERE`), state is derived from
  those members' own light entities instead — on/off, brightness, colour temperature
  and HS colour averaged the way core's `light.group` does — while control still goes
  out as one group frame.
- Never an auto-discovered entity for a group, area or general address: bus discovery
  keeps skipping broadcasts, exactly as before.

## How it's done

- **`custom_components/myhome/light_group.py`** (new): `MyHOMELightGroup(MyHOMEEntity, LightEntity)`.
  - Colour modes come from the same `dimmable`/`color_temp`/`rgb`/`hs` flags as a
    normal light (`_color_modes_from_flags`, shared with `light.py`).
  - `async_turn_on`/`async_turn_off` send `OWNLightingCommand.switch_on/off/
    set_brightness/set_color_temperature/set_hsv_color` against the group's own
    `#G` address; without members the sent values are applied optimistically
    (assumed-state semantics). A colour-only change carries the group's last known
    brightness as HSV "value" instead of zeroing it.
  - Subscribed to the gateway's `myhome_message_{mac}` signal: in assumed-state mode,
    any group event **or** the gateway's echo of a group dimension write
    (`*#1*#G*#D*…##`, which OWNd parses as a command, not an event) updates the
    entity; in member mode, group broadcast frames are ignored (members are the only
    source of truth there).
  - With `members`, resolves each declared `WHERE` to its light entity id via the
    entity registry and tracks their states (`async_track_state_change_event`),
    deriving on/off/brightness/colour-temp/HS the way `light.group` does.
  - `async_update()` and the initial poll-on-add skip the bus entirely when members
    are declared (nothing to ask the group for); an assumed-state group still polls
    its own status once on add and on demand.
  - `turn_on_timed` (the platform's custom service) raises a translated
    `HomeAssistantError` — a group has no native bus timer.
- **`custom_components/myhome/light.py`**: `build()` routes any `where` starting with
  `#` to `MyHOMELightGroup` instead of the normal `MyHOMELight`; a broadcast `where`
  that reaches here otherwise (general/area, matching `validate.py`'s own
  `General()`/`Area()` validators) is refused with a logged warning, never a light.
- **`custom_components/myhome/validate.py`**: `light_schema` gains an optional
  `members: [<WHERE>, ...]` key, validated as point-to-point addresses and only
  accepted when the light's own `where` is a group.
- **Docs**: README "YAML Configuration" gets a declared-groups section with the
  required wording (the group must already exist in the plant; declaring it here does
  not configure membership; use HA's own `light.group` to combine several lights
  instead), `known_limitations.md` and `ROADMAP.md` record what's covered and what's
  still open (a bus trace from a gateway that does **not** echo member status is still
  needed — see #368).

## How it's tested

- **`tests/test_issue_368_groups.py`** (new, 32 cases) drives real
  `OWNd.message.OWNMessage.parse()` frames against a `MyHOMELightGroup` built with a
  plain `AsyncMock` gateway (`send`/`send_status_request`), plus `PlatformDiscovery`
  through `async_setup_entry` for the yaml/registry build paths:
  - schema: members accepted on a group, rejected on point-to-point, rejected when
    not themselves a valid point-to-point address
  - `build()` creates a `MyHOMELightGroup` from `myhome.yaml` and from a restored
    registry entry; refuses an area/general `where` with a logged warning
  - assumed-state control: plain on/off, brightness, colour temperature, HS colour
    (checked against the exact frame string, e.g. `*#1*#6*#12*120*50*100##`),
    `transition` and `turn_on_timed` both raise
  - assumed-state bus updates: on/off, each dimension (1/12/14), the group
    dimension-write echo, a frame for a different group or WHO ignored, icon
    switching
  - member mode: resolution + tracking, mean brightness/colour-temp/HS, all-off
    clears derived values, a member reporting no attributes clears them too,
    unavailable members clear `available`, group broadcast frames ignored,
    unresolved member logged and skipped
  - polling: `async_update()`/poll-on-add skip the bus with members, poll it
    (status + brightness + colour-temp + HS) without
  - `tests/golden/frames/who01_lighting.yaml` gains the three group dimension-write
    frames (brightness/colour-temperature/HSV) verified against `openwebnet-mcp`'s
    grammar checker and regenerated into `corpus.json` /
    `test_golden_conformance.py`
- Full suite: **1654 passed**, **100% coverage** (including the new `light_group.py`),
  `ruff check` clean, `scripts/verify_ha_standards.py` clean.

## Not in scope

- Options-flow / panel editor for declaring a group (discussion #270) — `myhome.yaml`
  only for now.
- WHO=2 group/area/general handling for covers — own branches in `gateway.py`/`cover.py`.
- A trace from a gateway that does not echo member status after a group command
  (still missing per #368; the reactive re-sync side of P7 is PR #377).


Interstellar0verdrive:
Tested on hardware (not reviewed as code) on a MyHomeServer1 (firmware 2.87.13), HA 2026.9.2 in a container, a throwaway instance with nothing else configured, running this PR's head. I declared a group that really exists in my plant:

```yaml
light:
  bedroom_group:
    where: '#1'
    name: Group N
    members: ['33', '38']
```

**The light platform fails to set up:**

```text
08:44:19.309 ERROR [homeassistant.components.light] Error while setting up myhome platform
for light: 'MyHOMELightGroup' object has no attribute 'handle_event'
```

The traceback runs `light.py:188` → `discovery.py:459` → `:378` → `:397`. The group object is constructed (`light.py:108`) but never added to Home Assistant, and because the exception is raised inside `configure()` before `async_add_entities`, **no light entity is added at all** — not even the twenty restored from the registry. In that run the integration polled no light and processed no lighting frame; in the UI all twenty lights showed as unavailable.

Removing the declaration and restarting restores everything: no platform setup error, `light restored/configured 20 entities`, and the twenty lights are back (19 off, 1 on, matching the bus status frames). The only errors left in both runs are the usual `*#4*0##` / `*#16*0##` send retries on a plant without those subsystems, unrelated to this.

**Cause.** `discovery.py:397` subscribes every entity it creates with `entity.handle_event`:

```python
entity.async_on_remove(self.router.subscribe(self.who, keys, entity.handle_event))  # type: ignore[attr-defined]
```

`MyHOMELightGroup` doesn't define `handle_event`, and neither does `MyHOMEEntity`; the class listens on the dispatcher instead (`light_group.py:126-132`, handler at `:161`). The `type: ignore` on that line is presumably why nothing flagged it statically.

**Scope.** `build()` returns a group for any `where` starting with `#`, and `members` only changes `assumed_state`, so this hits every declared group with or without members, on any gateway — and after the first time, also on restore from the registry. Worth a regression test that sets up the platform with a declared group through `discovery`, since the current tests appear to exercise the class rather than the platform path.

The other platforms are unaffected: the cover platform set up its 12 entities normally in the same failing run.

Happy to re-test as soon as there's a commit — this rig has a real group with two members, so it can exercise both the assumed-state and the member-derived paths.

GreenGrassBlueOcean:
Hi @Interstellar0verdrive, thank you so much for the incredibly detailed test report! You hit the nail on the head—this was an architectural regression caused by MyHOMELightGroup missing the handle_event migration for the new V2 FrameRouter, and you are completely correct that the # type: ignore in discovery.py let it slip past our static analysis. 

We actually caught this exact crash in CI immediately after this PR merged. We've already implemented and merged the full fix in PR #391, which renamed _handle_bus_message to handle_event, removed the legacy dispatcher subscriptions, and—most importantly—added an abstract handle_event method to MyHOMEEntity while removing that # type: ignore, so mypy will strictly enforce this interface for all future entities.

If you pull the latest 2-phase1-architecture branch, your hardware test should now set up the platform perfectly and route the group frames natively through the FrameRouter! Thanks again for taking the time to test this on physical hardware!

Interstellar0verdrive:
Ha — I tested the PR's own head instead of the branch, so I managed to reproduce a bug you'd already fixed two hours earlier. Nothing like arriving at the scene of the crime after the police have left LOL. I'll pull the current `v2-phase1-architecture` and run the same test again; the rig has a real two-member group, so both the assumed-state and the member-derived paths can be exercised. Glad the diagnosis at least matched yours... 😆

Interstellar0verdrive:
Re-tested on `v2-phase1-architecture` at `180b759` (2026-09-17 06:51 UTC), which carries the #391 fix. Same rig: MyHomeServer1 (firmware 2.87.13), HA 2026.9.2 in a container, nothing else configured, and a group that really exists in the plant with exactly two member actuators.

**The platform sets up.** `light restored/configured 21 entities` at 09:08:46.454 — the 20 point-to-point lights restored from the registry plus the declared group. No platform setup error anywhere in the run; the only ERROR lines are two NACKed status requests for subsystems this plant doesn't have (`*#4*0##`, `*#16*0##`), unrelated to lights.

**Single-frame control works.** Turning the group on transmits exactly one frame, `*1*1*#1##` (sent 09:09:53.312); no per-member command is transmitted, and the two members echo their own status 68 ms and 78 ms later. Both lamps physically came on. Turning it off, after a second group-on at 09:11:46.287, transmits exactly one `*1*0*#1##` (09:11:50.283), with member echoes at +74 ms and +103 ms; both lamps went off.

**State follows the members.** With `members` declared, group broadcasts are ignored (`light_group.py:158`) and the state is `any(member is on)` (`light_group.py:224`). Switching member 33 off (status frame 09:10:59.831) left the group on; switching member 38 off (09:11:03.841) turned it off with that same state change. Entity states read in the UI, the frames read in the log.

One anecdote from having the plant in front of me, with nothing to do with Home Assistant or this PR, but it drove me slightly mad while testing: the physical group button's status LED follows **one** member and ignores the other. Left lamp on, LED orange; right lamp on alone, LED stays green; and it behaves that way whichever order you switch them in. I'll have to check whether that's something the group configuration lets you change in MyHOME_Suite, since it's plant programming rather than anything on the bus. Mentioning it only because it's a reminder of how little a group's apparent state can be trusted from outside: even the button that commands it doesn't track it.

GreenGrassBlueOcean:
Thanks so much for taking the time to re-test this on your physical hardware, @Interstellar0verdrive! It's great to have confirmation that the fix works perfectly in a real-world setup. Really appreciate the thorough verification!

PR #377, opened 2026-09-16
feat: debounced resync for group area general frames (#368 PR A)

GreenGrassBlueOcean:
This is a draft PR implementing the debounced resync sweep as discussed in #248 and tracked in #368.

### 🎯 The Goal
This PR implements a **"debounced reactive re-sync"** to ensure Home Assistant light entities stay correctly in sync when physical wall switches or scenes trigger MyHOME group, area, or general lighting commands, without flooding the bus with redundant traffic on gateways that already echo status updates.

### 🛠️ How it works
When a group (e.g., `*1*1*#6##`), area (e.g., `*1*0*3##`), or general (e.g., `*1*1*0##`) lighting command is detected:
1. **The Debounce Window (~250ms):** A short timer is started for that address.
2. **Smart Cancellation:** If the gateway spontaneously echoes individual point-to-point status frames during this window (like on F461/F454), the integration knows the gateway is handling it and cancels the sweep to save bus traffic.
3. **Targeted Sweeps:** If the bus stays silent, the integration proactively sends status requests:
   * **Groups:** Sweeps the group status (`*#1*#G##`).
   * **Areas:** Sweeps the area status directly (e.g., `*#1*3##`).
   * **General:** Instead of the forbidden general query (`*#1*0##`), it dynamically identifies which areas have registered lights in HA and sweeps those specific areas instead.
4. **Configurable:** Adds a new toggle in the integration's Options Flow (*"Sweep group/area/general light addresses for status after a debounced silence window"*), enabled by default, so users with lagging gateways can opt out.

### 🧪 How is it tested
- **Unit Tests:** `pytest` coverage added for the debouncing logic, registry area identification, and configuration toggle.
- **Golden Samples from Physical Hardware:** Connected to a physical MH200 gateway using the `myhome-gateway` MCP tool. We captured a live status sweep response burst (~100 frames), including interleaved automation (WHO=2) and gateway diagnostic (WHO=13) frames that the MH200 natively leaks during sweeps. A new test (`test_resync_golden_mh200_sweep_burst`) replays these raw frames to ensure the integration correctly ignores interleaved diagnostic events and never triggers infinite re-sync cascades from individual status echoes.
- **HA Standards:** Code passes the architectural standards script (`verify_ha_standards.py`) and is `ruff` clean.


Interstellar0verdrive:
The debounce reads only forward: `_resync_broadcast` skips the sweep when `self._last_point_frame > fired_at`, so only member echoes that arrive *after* the broadcast frame count. On my MyHomeServer1 they arrive about 0.9 s *before* it (the capture in #248), so the comparison is always false and every wall-switch press on a group would pay for a sweep that isn't needed. Comparing against a window on both sides — something like `fired_at - 1.5 s` as well as after — covers both orderings, and costs nothing on the gateways that echo late.

Two smaller things while you're in there:

- **250 ms is tight even for the forward case.** On @xtimmy86x's MyHomeServer1 the trailing echoes landed 130–200 ms after the group frame, so a slightly busier bus would push them past the window and sweep anyway. Half a second would still feel instant and leaves room for jitter.
- **`_last_point_frame` is global**, so any unrelated light reporting in the window cancels the sweep, and conversely a sweep can be triggered by silence on an address nobody touched. That's the best-effort heuristic @xtimmy86x described, which is fine, but it's worth a comment in the code saying so, and when declared members exist the check could be scoped to them.

Happy to run the result against my plant once it's out of draft: it's the ordering that the two MyHomeServer1 captures disagreed on, so it's worth confirming on both.

Issue #378, opened 2026-09-16
Regression of #254: the first command after ~30 s of silence is silently lost (MyHomeServer1)


Interstellar0verdrive:
**Regression: the idle-session fix from #254 is no longer in effect**

This is the same failure @marcob79 reported on an MH201 in #254 on 10 September, fixed the same day by `db36113` ("release idle command sessions"), which closed the command session proactively after `COMMAND_SESSION_IDLE_TIMEOUT` so nothing is ever written into a socket the gateway has already dropped.

On the current `v2-phase1-architecture` that release is gone: `COMMAND_SESSION_IDLE_TIMEOUT` is still defined but no longer referenced anywhere, and the sending worker awaits `self.send_buffer.get()` with no timeout. The session therefore stays open until the gateway kills it, which is exactly the pre-fix condition.

Measured below on a MyHomeServer1, on a quiet instance, where it reproduces every time.

Integration version: 2.0.0b12 (`v2-phase1-architecture` code, installed from #374's branch at `84c5e1a`)
OWNd: 2.0.0b6
Home Assistant: 2026.9.2, container, fresh instance
File usage: none, devices came from discovery only
Gateway: MyHomeServer1, firmware 2.87.13

**Any command sent after ~30 seconds of silence is lost, silently.**

The instance is a clean one I set up to look at the panel: one custom integration, no YAML, no automations, no dashboards, nothing else talking to the gateway. That turned out to be the condition that exposes this.

**Symptom.** Leave it idle for half a minute, then switch a light on from the UI: nothing happens, no error anywhere. Do it again a few seconds later and it works. It looks as if certain lights never respond, because a user presses *on* (lost), then presses *off* (delivered, but the light was never on, so nothing visible happens).

Measured by the gap since the previous successful command:

| Idle before the command | Result |
|---|---|
| 23 s | delivered |
| 36 s | lost |
| 48 s | lost |
| 87 s | lost |
| 159, 162, 163 s | lost |
| 244 s | lost |
| 377 s | lost |

So the gateway drops the command session after roughly 30 seconds, and the first command afterwards is always lost. Clicking around keeps the session warm, which is why it looks intermittent.

**The log, twice in a row on two different addresses:**

```text
*1*1*13##  successfully queued
*1*1*13##  successfully unqueued by worker 0
           Command session closed.
WARNING    Connection lost before acknowledgement of `*1*1*13##`; message will not be replayed.

*1*0*13##  successfully queued
[...]

Issue #254, opened 2026-09-10
MH201: sometime cover command seems to be lost


marcob79:
probably related to this error reported in the log, the command seems not be ack by the gateway.... maybe connection is not stable?
do you need any additional logs?

This error originated from a custom integration.

Logger: custom_components.myhome
Source: custom_components/myhome/ownd/connection.py:890
Integration: MyHome (documentation, issues)
First occurred: 08:05:21 (2 occurrences)
Last logged: 08:54:16

[MH201 gateway - 192.168.1.27] Connection lost before acknowledgement of *2*2*14##.
[MH201 gateway - 192.168.1.27] Connection lost before acknowledgement of *2*1*14##.

### MyHOME Diagnostic Bundle

**Environment:**
- **Home Assistant Version:** 2026.9.1
- **Integration Version:** 1.0.0-beta
- **Browser / User Agent:** Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36
- **Timestamp:** 2026-09-10T06:55:29.488Z

**Active Gateway Configuration:**
- **Model:** MH201 (BTicino S.p.A.)
- **Firmware:** 1.0
- **Connection:** Ethernet TCP (192.168.1.27:20000)
- **MAC Prefix:** 00:03:50
- **Queue Pacing:** 0.05s
- **Worker Count:** 1
- **Connection Status:** Connected

**Buffer Telemetry:**
- **Total RX Frames:** 57
- **Total TX Frames:** 11
- **Total Captured:** 68
- **Buffer Depth:** 50 / 200
- **Gateway Queue Depth:** 0
- **Active Card Filter:** None (All frames)

<details><summary>OpenWebNet Bus Trace</summary>

```
[06:05:23.318] [RX] *2*0*14##
[06:05:24.306] [TX] *2*1*14##
[06:05:24.377] [RX] *2*1*14##
[06:05:25.354] [TX] *2*0*14##
[06:05:25.428] [RX] *2*0*14##
[06:05:28.297] [TX] *2*1*14##
[06:05:28.377] [RX] *2*1*14##
[06:05:28.993] [TX] *2*0*14##
[06:05:29.067] [RX] *2*0*14##
[06:06:22.140] [RX] *#*0##
[06:07:52.141] [RX] *#*0##
[06:09:22.144] [RX] *#*0##
[06:10:52.145] [RX] *#*0##
[...]

PR #381, opened 2026-09-16
fix(gateway): close idle command sessions before the gateway drops them

GreenGrassBlueOcean:
## Goal

Fix #378: the first command sent after the gateway's own idle timeout (~30s, measured on a MyHomeServer1/MH200N) is silently lost. Nothing errors, no log entry, the service call returns successfully — the frame was written into a socket the gateway had already closed, and OWNd correctly declines to replay a written command, so it just disappears. A cover's stop frame taking this path leaves the shutter running to its end stop. Field reports from three different gateway families (MyHomeServer1, MH200N, MH201) confirm the same mechanism on different idle timelines (~30s to ~180s).

This is a regression of #254 (fixed by db36113): commit 267b794 hardened the sending worker's exception handling (a good fix — `send_buffer.task_done()` used to only run on the shutdown path, so an exception from `send()` could skip it and leak the socket) but, in the same diff, dropped the `asyncio.wait_for`/`TimeoutError` wrapping around `send_buffer.get()` that proactively closed an idle command session — and deleted both tests that covered it. `COMMAND_SESSION_IDLE_TIMEOUT` was left defined but never referenced again. Root-cause writeup: [#378 comment](https://github.com/OpenWebNet-HA/MyHOME/issues/378#issuecomment-5702622265).

## How

Restore the idle-timeout wrapping in `sending_loop` (`custom_components/myhome/gateway.py`), layered on top of 267b794's exception-safety structure rather than reverting it:

- `send_buffer.get()` is awaited under `asyncio.wait_for(timeout=COMMAND_SESSION_IDLE_TIMEOUT)`.
- On `TimeoutError` — the queue sat idle past the timeout — the command session is closed proactively via the existing `_session_is_open()` helper, then the loop continues waiting.
- The next real send already reconnects when `_session_is_open()` reports the streams gone (existing logic, untouched), so it opens a fresh session instead of writing into the one the gateway silently dropped.
- All of 267b794's hardening (per-task `try/except/finally`, `task_done()` always called, outer `try/finally` closing the session on worker exit, initial-`connect()` exception handling) is preserved as-is.

## Testing

**Unit + E2E coverage, both restored (deleted together in 267b794, so they come back together):**
- `test_sending_loop_idle_timeout_closes_session` — proves the idle socket gets closed. This alone was flagged in review as insufficient (@fedem95, @Interstellar0verdrive): it doesn't prove the *next* command still arrives.
- `test_issue_254_mh201_idle_disconnect_and_reconnection_e2e` — proves a command sent after the idle-close still reconnects and is delivered, which is the actual symptom every reporter on #378 saw as a lost command.

**Golden corpus, from the field data in the thread:**
- Added 3 fixtures to `tests/golden/frames/who02_automation.yaml` from @TheDarkWizard's [captured trace](https://github.com/OpenWebNet-HA/MyHOME/issues/378#issuecomment-5703741524): the WHO=2 command-translation echoes for STOP/UP/DOWN (`*2*1000#0|1|2*03##`). Validated grammar+semantics via `openwebnet-mcp` and confirmed OWNd parse/roundtrip before adding. `corpus.json` regenerated (110 → 113 frames), provenance recorded in `SOURCE.yaml`.
- These weren't in the corpus before — WHO=2 translation frames had zero coverage (only WHO=1 had any). Independent of the idle-timeout bug itself, but the same trace that proved the bug surfaced the gap.

**Results:**
- Full suite in this worktree, with both restored tests and the golden fixtures on disk: 1621 passed, 1 skipped. A separate full run in the main tree (before this branch existed) also passed clean: 1631 passed, 1 skipped.
- `tests/test_golden_conformance.py`: 269 passed, up from 263 (3 new frames, 2 tests each: parsing + roundtrip).
- `ruff check` clean on all changed files.

## How to pull this into `v2-phase1-architecture`

```bash
git fetch https://github.com/GreenGrassBlueOcean/MyHOME.git fix/issue-378-command-session-idle-timeout
git checkout -b fix-378-review FETCH_HEAD
python -m pytest tests/test_gateway.py -k "idle or issue_254" -v   # the two restored tests
python -m pytest tests/test_golden_conformance.py -q               # golden corpus, including the new fixtures
python -m pytest -q                                                 # full suite
```

To see the fix change actual behavior, flip `COMMAND_SESSION_IDLE_TIMEOUT` in `custom_components/myhome/gateway.py` down to a few seconds, point the integration at a real gateway, and leave it idle past that window before sending a command — the log should show `Command session idle for Ns; closing socket to release gateway resource.` followed by a normal reconnect-and-send, instead of the silent loss from #378.


custom_components/myhome/gateway.py
```diff
@@ -857,7 +857,25 @@ async def sending_loop(self, worker_id: int) -> None:
                 return
 
             while not self._terminate_sender:
-                task = await self.send_buffer.get()
+                try:
+                    task = await asyncio.wait_for(
+                        self.send_buffer.get(),
+                        timeout=COMMAND_SESSION_IDLE_TIMEOUT,
```
fedem95: Not a blocker for this PR, more a follow-up idea: 15 s fits the gateways with the shortest window (~30 s on MyHomeServer1/MH200N), but on the MH201 I measured the gateway closing idle command sessions only after ~150–180 s. With a single constant, every gateway reconnects on the first command after 15 s of silence (~50 ms on the MH201).

`GatewayProfile` in OWNd already carries per-model timing like `event_keepalive_interval`, and this loop already reads `profile.command_queue_delay`, so a `command_session_idle_timeout` on the profile would fit, with 15 s as the default for generic and unmeasured models. It would need an OWNd release first, so a separate issue is probably the right home for it.

One trade-off to keep in mind: closing early also frees a session slot sooner, which matters on gateways with few slots or with more than one client connected.

GreenGrassBlueOcean: Great idea @fedem95 — we went ahead and made the integration forward-compatible in e43397e!

`MyHOMEGatewayHandler` now exposes a `command_session_idle_timeout` property that checks `getattr(self.gateway.profile, "command_session_idle_timeout", None)` and falls back to `COMMAND_SESSION_IDLE_TIMEOUT` (15.0s). In `sending_loop`, the wait timeout evaluates `self.command_session_idle_timeout` dynamically on each iteration so that as soon as a gateway model is identified (or once OWNd ships `command_session_idle_timeout` on `GatewayProfile`), the custom profile timeout kicks in immediately.

Also completely agree on the slot-conservation trade-off: on multi-client setups or slot-limited hardware (and as @TheDarkWizard's trace neatly illustrated with 4 concurrent workers leaving all 4 sessions in `CLOSE-WAIT`), releasing after 15s keeps the pool healthy for all clients without waiting out the full ~150–180s on the gateway side.

tests/golden/SOURCE.yaml
```diff
@@ -92,6 +92,18 @@ authorities:
...
         devices: "Thermoregulation 3550 central unit (#0), 6 zones (LN4691), F430R8 actuators"
+      - name: "Physical Bus Capture (BTicino MyHomeServer1 + Automation WHO 2, idle command-session loss)"
+        reference: "https://github.com/OpenWebNet-HA/MyHOME/issues/378#issuecomment-5703741524"
+        author: "TheDarkWizard"
+        date: "2026-09-16"
+        gateway: "BTicino MyHomeServer1"
+        devices: "Cover/shutter actuator, address 03; WHO 2 command-translation echoes (*2*1000#0/1/2*03##)"
+      - name: "Physical Bus Capture (BTicino MH201 + Lighting WHO 1, idle command-session test)"
+        reference: "https://github.com/OpenWebNet-HA/MyHOME/issues/378#issuecomment-5704228050"
+        author: "Federico Manzella (@fedem95)"
```
fedem95: @GreenGrassBlueOcean small request on 23e00de: could the `SOURCE.yaml` entry credit just `@fedem95` instead of my full name? Thanks.

Interstellar0verdrive:
Tested on a MyHomeServer1 (firmware 2.87.13), one command worker, a throwaway Home Assistant instance with nothing else configured.

**Without this PR** — same instance earlier the same evening, 19:19 to 20:32: 52 commands and 37 status requests delivered, and **18 commands lost**. Status requests were never lost, since OWNd replays those. Measuring the silence from the previous attempt on the command session:

- every loss followed at least **36.1 s** of silence (minimum 36.07 s);
- every delivery followed at most **29.2 s**.

**With this PR** — 23:52 to 00:15: 6 commands and 33 status requests, **nothing lost**. The commands that followed a long silence, in order:

| Silence before the command | Result |
|---|---|
| 78.0 s | delivered, first attempt |
| 67.7 s | delivered, first attempt |
| 115.4 s | delivered, first attempt (a cover) |
| 115.2 s | delivered, first attempt (a cover) |
| 999.5 s | delivered, first attempt |

Each of the five is preceded in the log by `Command session idle for 15.0s; closing socket to release gateway resource`, so the socket the gateway would have dropped is never the one we write into. Reopening costs 11–13 ms for the handshake, and the frame leaves 45–111 ms after the reopen begins.

Two honest caveats. The "after" side is five post-idle commands against eighteen losses before, so it shows the failure gone rather than proving a rate. And the logs alone don't establish that nothing else changed: both runs report the same integration version string, since this branch doesn't bump it, so the distinction between the two builds rests on my having installed this branch, not on anything in the log.

Happy to run a longer session if that helps — this plant reproduced the failure every time before the fix, so it's a cheap place to test.

GreenGrassBlueOcean:
Thanks for the detailed test results! Let's include these tests as a golden sample. We'll plan to do further stress testing during the beta release.

Issue #371, opened 2026-09-16
Sending worker passes retry_after_lost_ack to OWNCommandSession.send(); no frame reaches the bus


GreenGrassBlueOcean:
### Summary

On `v2-phase1-architecture` (since `267b794`, landed through #361 on 2026-09-16) the sending worker calls

```python
collected = await _command_session.send(
    message=task["message"],
    is_status_request=task["is_status_request"],
    retry_after_lost_ack=True,
)
```

([gateway.py:890-894](https://github.com/OpenWebNet-HA/MyHOME/blob/3f5e791/custom_components/myhome/gateway.py#L890-L894)). No OWNd release or branch defines `retry_after_lost_ack` — `OWNCommandSession.send(self, message, is_status_request=False)` in 2.0.0b6, on `master`, and on #35/#36:

```
>>> OWNCommandSession.send(message="*1*1*11##", is_status_request=False, retry_after_lost_ack=True)
TypeError: OWNCommandSession.send() got an unexpected keyword argument 'retry_after_lost_ack'
```

### Impact

The `TypeError` is raised inside the worker's `try`, caught by its `except Exception`, logged as *"unexpected error while sending `…`; delivery is unconfirmed"*, and the frame's delivery future is cancelled. **Every** command and status request goes this way, so the integration receives events but can never write to the bus. Not in 2.0.0b11; any beta cut from the current head would ship it.

### Why CI is green

The four tests covering the worker replace `OWNCommandSession` with a `MagicMock` whose fake `send` *accepts* the keyword (`tests/test_gateway.py:739`, `:777`, `:803`; `tests/test_issue_315.py:184`; `tests/test_phase1_architecture.py:520`). Nothing runs the worker against the real class.

### Fix

- Drop the keyword. OWNd's `_locked_send()` already owns the retry policy: a status request may be retried once after a transport reset; a written command is never replayed.
- Add `tests/test_command_session_contract.py`: the worker driving the **real** `OWNCommandSession` over TCP to `MockGatewayHarness`, plus an `autospec` variant that enforces OWNd's signature without a socket. Both fail on the current head, both pass with the fix.
- While here: `tests/test_connection.py::test_get_next_heartbeat_timeout` asserted that `close()` is the method that recycles the socket; OWNd #35 routes that through `_close_streams()`. The test now asserts the outcome (streams released, reconnect attempted) and passes on b6 and on #35/#36.

[...]

PR #382, opened 2026-09-16
ci: disable snapshot auto-update and enforce command session kwargs lint

GreenGrassBlueOcean:
## Summary & Motivation

This PR hardens CI and introduces static contract validation for the integration's OpenWebNet command transport. It addresses two critical blind spots in our test harness:

1. **Static AST signature enforcement for `OWNCommandSession.send`**
2. **Disabling snapshot auto-update and auto-commit in CI**

---

### 1. The #371 Outage: Mocks Swallowed a Fatal `TypeError`
In commit `267b794` (landed via #361), the sending worker was updated to pass `retry_after_lost_ack=True` to `_command_session.send()`:

```python
collected = await _command_session.send(
    message=task["message"],
    is_status_request=task["is_status_request"],
    retry_after_lost_ack=True,
)
```

No version of OWNd (neither `2.0.0b6`, `master`, nor upcoming branches) defines `retry_after_lost_ack`. The call raised `TypeError: OWNCommandSession.send() got an unexpected keyword argument 'retry_after_lost_ack'`.

Inside the worker loop, this exception was caught by a broad `except Exception`, logged as an *"unexpected error while sending"*, and the frame's delivery future was cancelled. **Every command (lighting, covers, climate) and status request on every physical gateway failed to reach the bus.**

**Why CI stayed green:**
All unit tests exercising the sending worker (`test_gateway.py:739`, `:777`, `:803`; `test_issue_315.py:184`; `test_phase1_architecture.py:520`) replaced `OWNCommandSession` with a bare `MagicMock()`. Because vanilla mocks accept arbitrary keyword arguments, the bad call was silently swallowed in CI and was only caught when deployed to live hardware.

### 2. Silent Snapshot Mutation in CI
In `.github/workflows/test-coverage.yaml` (line 39) and `pypi_standards.yml` (line 83), pytest was invoked with the `--snapshot-update` flag. Furthermore, `test-coverage.yaml` automatically staged and committed `tests/**/__snapshots__/*` back to the default branch on push:

```yaml
PYTHONPATH=. pytest tests/ --snapshot-update ...
git add "tests/**/__snapshots__/*" || true
git commit -m "chore: auto-generate coverage badge & README table [skip ci]"
```

When CI auto-updates snapshots, snapshot assertions stop acting as regression barriers. If code changes unintentionally alter serialized output, diagnostics, or state dictionaries, the test run silently overwrites the snapshot files and commits the regression rather than failing the build.

---

### 1. Static AST Kwargs Audit (`test_all_command_session_send_calls_match_ownd_signature`)
We added an AST-based audit to `tests/test_command_session_contract.py`. Rather than relying solely on test coverage or mock fidelity:
- It dynamically inspects the true signature of `OWNCommandSession.send` via `inspect.signature(OWNCommandSession.send).parameters`.
- It parses every `.py` file under `custom_components/myhome/` with `ast.parse`.
- It checks every `.send()` call targeting `command_session` instances and asserts that all passed keyword arguments are strictly within the documented signature, and that positional arguments do not exceed the limit.

**Why this works:**
- **Mock-proof**: It analyzes source AST before mocks even exist.
- **Path-proof**: It verifies all call sites across the codebase (including `custom_components/myhome/core/transport/tcp.py` and `gateway.py`), even code inside rare error-recovery branches that unit tests might not exercise.
- **Fail-fast**: Any invalid kwarg immediately fails the test suite at build/test time with the exact file and line number.

### 2. Deterministic, Read-Only Snapshots in CI
- Removed `--snapshot-update` from both `test-coverage.yaml` and `pypi_standards.yml`.
- Removed `git add tests/**/__snapshots__/*` from the CI commit script.

**Why this works:**
- Snapshots in CI are now strictly read-only assertions. Any unintended deviation between code output and committed snapshots causes an immediate, loud CI failure.
- Snapshots may only be regenerated locally by a developer using `pytest --snapshot-update`, requiring the diff to be reviewed and intentionally committed.

---

## Verification

- **WSL (Python 3.14.7 / Home Assistant Core 2026.9.2)**:
  ```bash
  pytest tests/test_command_session_contract.py -v
  ```
  `4 passed in 2.22s` (TCP harness test, autospec mock test, golden MH200 test, and the new AST static audit).
- **Static code hygiene**:
  ```bash
  ruff check custom_components tests scripts
  ```
  `All checks passed!`


Interstellar0verdrive (approved):
Approving: removing `--snapshot-update` from CI is the part I'd have asked for anyway — a suite that rewrites its own expectations when behaviour changes, and commits them, is not a regression barrier. Dropping the snapshot auto-commit goes with it.

The AST audit is a good shape for this class of problem, and I like that it asserts it found at least one call, so it can't pass by looking at nothing.

One limit, not a blocker: it recognises a call only when the receiver's name contains `command_session`, so a rename or a call through a local alias silently drops out of the audit while the test keeps passing. Matching on the type rather than the name, or asserting the number of audited call sites, would keep it honest through refactors — and this is a file whose whole point is surviving refactors.

Issue #380, opened 2026-09-16
A cover's position advances even when its direction frame never reached the bus


Interstellar0verdrive:
Integration version: 2.0.0b12 (`v2-phase1-architecture` code, installed from #374's branch at `84c5e1a`)
OWNd: 2.0.0b6
Home Assistant: 2026.9.2, container, fresh instance
File usage: none, devices came from discovery only
Gateway: MyHomeServer1, firmware 2.87.13

**A shutter that never moved is reported at a position it never reached.**

Sequence, all of it from one `cover.set_cover_position` to 25 % on a cover that was fully open:

```text
20:22:21.718  *2*2*91##  queued
20:22:21.719  *2*2*91##  unqueued by worker 0
20:22:21.722  WARNING    Connection lost before acknowledgement of `*2*2*91##`; message will not be replayed.

20:22:41.976  *2*0*91##  queued
20:22:42.028  RX         *2*1000#0*91##
20:22:42.078  *2*0*91##  successfully sent
20:22:42.092  RX         *2*0*91##
```

The direction frame was never delivered (that loss is #378). Nothing else was sent, and the shutter did not move: I watched it, it stayed fully open for the whole twenty seconds. Yet the entity afterwards reads:

```json
{
  "state": "open",
  "current_position": 25,
  "last_run_direction": "close",
  "last_run_seconds": 18.76,
  "last_run_ended_at": "2026-09-16T18:22:42.079+00:00"
}
```

So the integration timed a full 18.76 s run, wrote the stop that ends it, and moved its estimate to the requested position, for a movement that never happened.

**Where it comes from.** `_await_motion_anchor` treats a cancelled delivery as "no echo to expect" and falls through to anchoring on *now* (`cover.py:544-548`):

```python
except asyncio.CancelledError:
    if not written.cancelled():
        raise  # our own task was cancelled (new command, entity removed): stop here
    # the *delivery* was cancelled (queue flushed): no echo, anchor on "now" below
    self._end_echo_window()
```

The assumption behind the comment is that a cancelled delivery means the frame went out and the queue was flushed. It also covers the case where the frame was never written at all, and then the run that follows is fiction.

**Why it matters.** The position of a timed cover is the only thing the integration knows about it, and here it becomes wrong in a way nothing can correct: every later relative movement builds on it, an automation that "closes the covers" leaves one open while Home Assistant reports it closed, and the user has no signal at all — `cover.set_cover_position` returned successfully.

[...]

PR #385, opened 2026-09-16
fix(cover): abort motion and prevent position advance on undelivered frames (#380)

GreenGrassBlueOcean:
## Summary

Resolves #380: *"A cover's position advances even when its direction frame never reached the bus"*.

### Root Cause & Regression Background
In #302, timed covers were anchored on physical frame writes. However, `_await_motion_anchor` caught `asyncio.CancelledError` and `TimeoutError` and fell through to:
```python
if self._move_start_time is None:
    self._anchor_run(time.monotonic())
```
When commit `267b794` dropped `COMMAND_SESSION_IDLE_TIMEOUT` (#378), idle command sessions remained open for hours. Any subsequent direction command encountering a stale socket caused delivery futures to be cancelled upon queue flush. Because of the fallback, `_auto_stop()` accepted `time.monotonic()` as an anchor, waited for the run duration (e.g. 18.76 s), advanced the Home Assistant state to the requested position (e.g. 25%), and queued a STOP command `*2*0*where##` for a movement that never took place.

Additionally, `_track_write` on `fut.cancelled()` only closed the echo window, leaving `_attr_is_opening` / `_attr_is_closing` stuck as `True`.

### Proposed Changes
1. **`_abort_motion()` Helper**:
   - Resets `_move_start_time`, `_run_started_at`, `_motion_started_at`.
   - Clears `_attr_is_opening = False` and `_attr_is_closing = False`.
   - Restores `_start_position = self._attr_current_cover_position` and sets `_attr_is_closed = (pos == 0)`.
   - Cancels scheduled `_stop_task` and ends the echo window.
   - Publishes the updated state to Home Assistant.
2. **`_await_motion_anchor()`**:
   - When delivery future is cancelled, times out (`WRITE_TIMEOUT`), or raises an exception: calls `_abort_motion()` and raises `HomeAssistantError` with translation key (`command_delivery_cancelled`, `command_delivery_timeout`, `command_delivery_failed`).
3. **`_auto_stop()`**:
   - Catches `HomeAssistantError`, logs the abort, and exits cleanly without sleeping, without setting `_attr_current_cover_position = target_position`, and without dispatching a STOP frame.
4. **`_track_write()`**:
   - Calls `_abort_motion()` when a direction command future is cancelled or raises an exception.
5. **`_async_move()`**:
   - Immediately aborts and raises `HomeAssistantError` if the returned delivery future was already cancelled or failed.
6. **Golden Sample Tests**:
   - Uses physical MH200 plant fixtures (`tests/fixtures/plants/mh200_physical_plant/myhome.yaml`):
     - `Cover 11I02` (`where: "11"`, `interface: "02"` -> OpenWebNet address `11#4#02`)
     - `Cover 85` (`where: "85"`)
   - Verifies exact Issue #380 sequence: position remains unchanged, is_opening/closing revert to False, and no false STOP frame is sent.
   - Verifies calibration aborts immediately without hanging 60 seconds when direction delivery fails.
   - Updates `tests/test_issue_302.py` tests that previously asserted the invalid fallback.

### Verification
- `pytest tests/test_component_cover.py tests/test_issue_302.py tests/test_cover_calibration.py tests/test_issue_380_cover_delivery.py`: **87 passed** in WSL Python 3.14.
- `scripts/verify_ha_standards.py`: **All architectural standards passed**.
- `ruff check custom_components tests scripts`: **All checks passed**.

**Note:** As designed in this PR, the caller of the cover service only sees the error when the delivery future is already resolved at call time. When the future resolves a few milliseconds later, `_auto_stop` catches it and logs an abort, so `cover.close_cover` and `cover.set_cover_position` still return successfully, which means an automation closing the covers cannot tell that nothing actually moved.

### Design Note: Fire-and-Forget vs. Blocking Service Calls

As designed in this PR, the caller of the cover service only sees an error if the delivery future is *already* resolved/cancelled at the exact moment of the call. If the future is cancelled a few milliseconds later (e.g., due to a queue flush), the background `_auto_stop` task catches the failure and logs an abort, but the original service call (e.g., `cover.close_cover`) still returns successfully.

**The Case In Favor (Current Design):**
- **Responsiveness:** Returning immediately after enqueuing prevents the UI or caller from freezing while waiting for physical bus confirmation, which could take hundreds of milliseconds on a congested bus.
- **Prevents Automation Stalls:** Automations aren't bottlenecked by bus latency and can proceed without waiting for the gateway's write queue to flush.
- **Optimistic State Alignment:** Home Assistant naturally favors optimistic state changes. Queuing the intent and letting a background task reconcile the physical result later aligns well with this paradigm.

**The Case Against (Why blocking could be better):**
- **Silent Automation Failures:** An automation (e.g., "Close shutters, then arm alarm") assumes the cover moved because the service call returned success. It remains blind to the fact that the frame was dropped.
- **Breaks Native Error Handling:** Users cannot leverage HA's native `catch` blocks, `continue_on_error`, or retry loops because the exception is swallowed into an `Auto-stop aborted` log rather than bubbling up to the automation trace.
- **Misleading Feedback:** Manual UI triggers appear to succeed instantly without any error toast if the command fails moments later.


Interstellar0verdrive:
Tested on hardware rather than reviewed as code: a MyHomeServer1 (firmware 2.87.13), one command worker, a throwaway instance with nothing else configured, running this PR's head. Its base doesn't include #381, so the undelivered-frame condition still reproduces on demand after about half a minute of silence. The cover under test has a real travel of 14.2 s and runs on the default 25 s configured travel.

**Undelivered frame, `close_cover`** (00:35:51, after 95 s of silence):

```text
00:35:51.737  *2*2*91##  queued
00:35:51.740  WARNING  Connection lost before acknowledgement of `*2*2*91##`; message will not be replayed
```

No `*2*0*91##` is written between then and the next command at 00:37:26, so no stop frame followed. The entity stayed at position 100 and the shutter did not move (both read from the UI and from the window — the log doesn't carry entity state).

**Undelivered frame, `set_cover_position` to 60 %** (00:38:52, after 66 s of silence):

```text
00:38:52.912  *2*1*91##  queued
00:38:52.916  WARNING  Connection lost before acknowledgement of `*2*1*91##`; message will not be replayed
00:38:52.917  ERROR    Auto-stop aborted for 91: Cover 91: direction command delivery was cancelled before reaching the bus
```

Again no stop frame, position unchanged at 25, shutter still closed.

**Delivered frame, `set_cover_position` to 25 %** (00:37:26): motor-start echo at 00:37:27.48, stop queued at 00:37:46.24 and sent at .278 — an 18.75 s run, exactly 75 % of the configured travel. Normal behaviour intact.

For contrast, the same rig before the fix, from #380: `*2*2*91##` lost at 20:22:21.72, a 20.3 s timed run anyway, a stop actually written at 20:22:42.08, and the position advanced to the target for a movement that never happened.

One note rather than a request, since the PR doesn't claim otherwise: the caller only sees the error when the delivery future is already resolved at call time. When it resolves a few milliseconds later, as in both episodes above, `_auto_stop` catches it by design and it becomes the `Auto-stop aborted` log line, so `cover.close_cover` and `cover.set_cover_position` still return successfully. Worth a line in the PR description, because an automation that closes the covers can't tell that nothing moved.

PR #387, opened 2026-09-16
feat(gateway): generate repair issue for unconfigured timezone (999)

GreenGrassBlueOcean:
### Goal
Surfaces the "Unconfigured Timezone" (999 sentinel) edge-case for MyHOME gateways (such as MH200N, F454, or MyHomeServer1) via the Home Assistant Repairs framework instead of silently failing or dropping the connection.

### Why this happens
OpenWebNet gateways default their timezone to a placeholder string `"999"` when the user hasn't explicitly configured a timezone in the gateway's web interface or software suite (TiMyHome / MyHOME_Suite).
When the Home Assistant integration queries the gateway for its Date & Time (`WHO=13`, Dimension `0` or Dimension `22`), the gateway responds with `*#13**0*<HH>*<MM>*<SS>*999##` or `*#13**22*<HH>*<MM>*<SS>*999*...##`.
The underlying `OWNd` protocol library expects a standard timezone offset here, causing an `Invalid gateway timezone: '999'` exception that drops the connection.

### How it is fixed
While the underlying python `OWNd` parser implements a compatibility shim to intercept the `999` and prevent crashes by converting it to `""` (empty/unknown), this effectively hides the configuration error from the user.
This PR introduces a check inside `gateway.py` (`_handle_gateway_diagnostics`) that specifically watches for `999` across both Dimension 0 and Dimension 22:
- When detected, it generates a user-actionable repair issue in Home Assistant Settings with clear instructions and a link to the documentation in `docs/configuration/gateways.md#gateway-timezone-configuration`.
- When a valid timezone offset is received (e.g. `+01:00`), the repair issue automatically resolves and clears.

### Why this is the "Home Assistant way"
According to Home Assistant's Architectural Standards, integrations should fail gracefully and rely on the **Issue Registry / Repairs Framework** for actionable user configuration errors, rather than spamming system logs (which normal users rarely read). By generating a Repair Issue (`unconfigured_timezone`), we alert the user directly on their dashboard that their physical gateway requires attention, providing clear instructions on how to solve it without breaking the rest of their setup.

### Testing
- Validated via `pytest` (1,616 passed, 100% test coverage) and `ruff` locally on the Python 3.14 WSL environment.
- Verified Dimension 0 and Dimension 22 frames with `999` sentinel and valid offsets.
- Verified integration translations across `strings.json`, `en.json`, `it.json`, `fr.json`, and `nl.json`.
- `scripts/verify_ha_standards.py` passed with 0 violations.

Related to #305


GreenGrassBlueOcean:
@xtimmy86x is maybe this (and it is upcoming generalization)  in https://github.com/OpenWebNet-HA/MyHOME/pull/388 something for the ui?

xtimmy86x:
Thanks, the overall approach makes sense to me. Detecting the raw `999` value, keeping the connection alive, creating an HA Repair, and automatically removing it once a valid timezone is received is the correct lifecycle.

Before approval, I found a few user-visible details that should be corrected:

1. The description contains double-escaped newlines. After parsing `strings.json`, the value still contains literal `\n\n`, so Home Assistant may display those characters instead of actual line breaks.

2. The `learn_more_url` currently points to `https://github.com/OpenWebNet-HA/MyHOME/wiki/Configuration#timezone`, but that wiki page/section does not appear to exist. We should either add the documentation first or point it to an existing page under `docs/configuration/`.

3. The Repair description says that an unconfigured gateway timezone causes incorrect Bus Monitor timestamps. However, the Bus Monitor timestamps are generated from Home Assistant's UTC clock and rendered in the browser's local timezone since #343. The `999` value causes an invalid gateway date/time response and parser failure, rather than directly changing the Bus Monitor timestamps. I would make the wording more precise.

4. The new entries in the Italian, French, and Dutch translation files are still in English. This is not a functional blocker by itself, but it would be better either to provide the translations or leave their normal translation workflow untouched.

Once these small points are addressed, the implementation looks good to me and should be ready for approval.

xtimmy86x:
> @xtimmy86x is maybe this (and it is upcoming generalization) in #388 something for the ui?

Yes, I think these warnings would be useful in the dedicated panel, especially on the gateway overview page.

My preference would be to keep the Home Assistant Repairs registry as the authoritative source and use the panel only as an additional presentation layer. For example, the panel could show:

* an active gateway warning such as “Timezone not configured”;
* the raw unknown WHO=13 model code;
* a link to the corresponding Home Assistant Repair;
* actions to export diagnostics or open the issue tracker when appropriate.

I would not duplicate the detection or Repair lifecycle in frontend code, since that would create a second source of truth. The backend should continue creating and resolving Repairs independently, while the v2.1 panel can make the same information easier to discover and act upon.

I also would not make #387 or #388 depend on the panel work: the Repairs should remain complete and useful for users who never install or open the dedicated panel.

fedem95 (commented):
Agree with xtimmy's point!
Two small extras:
- only dimension 0 is checked. Dimension 22 carries the timezone in the same field, and *#13**22*...*999*...## parses fine but never raises or clears the repair.
- the log in #305 comes from a MyHomeServer1, not only MH200N/F454, so the text could just say "the gateway".

Rest looks good, repair lifecycle works and it merges clean on the current base.

GreenGrassBlueOcean:
Thanks for the reviews @xtimmy86x and @fedem95! All points have been addressed in commit f497a68:

1. **Dimension 22 support**: Extended `_handle_gateway_diagnostics` to check `dim in (0, 22)`, so both dimension 0 and dimension 22 date/time frames raise or resolve the repair issue when encountering `999` or a valid timezone offset.
2. **Double-escaped newlines**: Corrected `\n\n` in `strings.json` and all translation files to ensure proper line breaks render in Home Assistant.
3. **Documentation & Learn More URL**: Added the `## 🕒 Gateway Timezone Configuration` section to `docs/configuration/gateways.md` and updated `learn_more_url` to link directly to it.
4. **Repair wording & hardware scope**: Clarified that an unconfigured timezone causes gateway date/time parsing failures and dropped diagnostic telemetry (rather than Bus Monitor timestamp corruption), and generalized references to "the gateway".
5. **Localization**: Provided native translations across `it.json`, `fr.json`, and `nl.json`.
6. **Tests**: Expanded test suite to cover Dimension 22 frames with 100% test coverage.

GreenGrassBlueOcean:
@xtimmy86x @fedem95 All review items have been addressed in commit f497a68 and the PR description has been updated accordingly.

Could you please take another look and re-approve when you have a moment? Thanks!

xtimmy86x:
Thanks, all the review points regarding Dimensions 0 and 22, wording, line breaks, translations, documentation and tests have been addressed correctly.

I found just one remaining detail before approval: the Repair currently links to:

`https://github.com/OpenWebNet-HA/MyHOME/blob/master/docs/configuration/gateways.md#gateway-timezone-configuration`

The new documentation does not currently exist on `master`, so the link returns a 404 for users testing or running the beta branch. Also, because the heading contains the `🕒` emoji, GitHub may generate a different anchor.

Could we remove the emoji from the heading and either point the link to `v2-phase1-architecture`, or otherwise ensure that the target exists on `master` before this code is released?

Once that final link is corrected, this looks ready for approval to me.

GreenGrassBlueOcean:
Thanks @xtimmy86x! Both points have been addressed:

1. **Heading Anchor**: Removed the `🕒` emoji in `docs/configuration/gateways.md` so the heading is `## Gateway Timezone Configuration`, ensuring GitHub deterministically generates `#gateway-timezone-configuration`.
2. **Permanent Documentation Link**: Instead of pointing to a temporary branch name that would need to be changed back upon release, we published a permanent, branch-independent guide on the GitHub Wiki:
   https://github.com/OpenWebNet-HA/MyHOME/wiki/Gateway-Timezone-Configuration

   The page is live, linked in the v2.0 Beta Hub and Sidebar, and covers WHO=13 RTC telemetry (`999` sentinel), resolution across Web UI & MyHOME Suite, and automatic repair clearance in Home Assistant.

The repair issue and tests have been updated accordingly, and the full test suite (100% coverage) and architectural standards are passing.

xtimmy86x:
All is addressed and ok for me, approved

PR #389, opened 2026-09-16
fix(light): migrate declared groups to v2 router architecture completely

GreenGrassBlueOcean:
Fixes the architectural routing breakage caused when PR #376 was merged. MyHOMELightGroup was still using _handle_bus_message and manual dispatcher subscriptions instead of exposing handle_event for the v2 Router. All tests pass locally.


GreenGrassBlueOcean:
Thanks so much @fedem95 for looking into this and solving the problem! Apologies for seeing this a bit too late—we ended up implementing this same fix in another PR that just got merged, so I'm closing this one out. Really appreciate the help though!

PR #411, opened 2026-09-18
docs(covers): clarify automated travel time calibration sequence

GreenGrassBlueOcean:
### Summary

Updates and clarifies the documentation for the **Calibrate Travel Time** button entity (`button.<name>_calibrate_travel_time`) in `docs/configuration/covers.md`.

### Context & Motivation

Previously, `docs/configuration/covers.md` under **Method 1: Stopwatch Button on Device Page** described the button as a manual toggle (*"click to start closing... when it reaches the bottom, click again to lock in the time"*). This was inaccurate and confusing for users and testers:
- The entity button directly calls `myhome.calibrate_cover`, which runs an automated 3-step sequence on the OpenWebNet bus.
- Clicking the button a second time while running actually triggers a `calibration_in_progress` error.
- Users and testers were anxious about pressing the button without knowing what the physical shutters would do or if they had to time/stop it manually.

### Changes
- Renamed Method 1 to **Automated Calibration Button on Device Page**.
- Documented the exact 3-step automated cycle:
  1. **Full Open (Sync)**: Drives to the top mechanical limit to establish a reliable reference position.
  2. **Full Close (Timed)**: Drives down, timing motor start to actuator stop on the bus (`travel_time_down`).
  3. **Full Open (Timed)**: Drives back up to the top limit, timing motor start to actuator stop (`travel_time_up`).
  4. **Saved Automatically**: Persisted into config entry options and cover attributes; shutter ends at 100% open.
- Added explicit notes:
  - Do not press the button a second time.
  - How to abort at any time (physical switch, HA stop button, or `myhome.stop_cover_calibration`).
  - Explanation of the 60-second actuator cutoff safety check, guiding users to Method 3 (Lovelace card stopwatch) or Method 4 (service action) if their actuator has unconfigured potentiometer limits.


Interstellar0verdrive:
Thanks for asking before merging. Dropping "press again" is right — the second press only raises an error. Two things in the new text would still send people into the problem I described in #409, though:

- "Abort with the HA stop button": as the code stands, a stop written by the integration does not abort the run, it ends it and the elapsed time is saved as the measurement. That is exactly how my 5.7 s came about. Until that is fixed I'd say the opposite — do not press Stop during calibration — and offer only `myhome.stop_cover_calibration` (or the wall switch) as the way out.
- Method 3 (the card's stopwatch) isn't in b13: #355 was closed unmerged and the shipped card has no stopwatch. Pointing 60 s-cutoff users there leads nowhere; the working alternative today is `myhome.set_cover_travel_time`.

And I'd open the section with the one thing a reader needs first: this works only if the actuator reports its stop when the shutter actually reaches the end stop. If the first run ends at about 60 s, the actuator can't be measured over the bus and the button will always refuse.

GreenGrassBlueOcean:
Thank you @Interstellar0verdrive for the detailed and spot-on review!

We have updated the PR in commit 84b2e69 to incorporate all your feedback:
1. **Actuator Prerequisite Upfront**: Added an `[!IMPORTANT]` callout right at the top of the calibration section explaining that automated on-bus calibration requires the actuator to emit a stop frame upon reaching the end stop. Actuators left at the default 60 s hardware cutoff will be safely rejected by the integration, directing users to `myhome.set_cover_travel_time`.
2. **HA Stop Button Warning**: Replaced the previous advice with an explicit warning: **Do NOT press the standard Stop button on the cover entity in Home Assistant during calibration**, explaining that doing so causes the integration to interpret the resulting stop frame as end-of-run and store the partial elapsed duration. Clarified that the safe ways out without saving are using a physical MyHOME wall switch or calling `myhome.stop_cover_calibration`.
3. **Removed Unmerged Card Stopwatch**: Removed Method 3 (Lovelace card stopwatch) since #355 was closed unmerged. Promoted manual entry via `myhome.set_cover_travel_time` (with a handheld stopwatch) as Method 3. Also purged ghost references to the card stopwatch across `known_limitations.md`, `services.md`, `use_cases.md`, and `troubleshooting.md`.

*(Note: We can also track a separate code bugfix for `async_stop_cover()` in `cover.py` to ensure calling stop on the entity during calibration explicitly interrupts the run via `_interrupt_calibration()` rather than recording premature run times.)*

Interstellar0verdrive:
Thanks — that's exactly the direction, and quick.

One correction before it merges, on the ways out. `myhome.stop_cover_calibration` is safe: it marks the run as interrupted and stores nothing. The wall switch is not. I just checked on my plant: stopping the shutter from the wall, with either button and in either direction, puts a plain stop frame on the bus — the very same frame as the actuator's own end-stop — and the integration treats it the same way: end of run, duration stored. Only an explicit open/close command from the wall is recognised as an interruption, and a stop press doesn't send one. So I'd word it as: don't stop the shutter at all during calibration, neither from the entity nor from the wall; the one safe way out is `myhome.stop_cover_calibration`.

On the code fix: yes, please track it. Two halves, though. A stop sent from the entity can be recognised and turned into an interruption. A stop coming from the bus cannot, because it is indistinguishable from the end stop — for that case the only defence is a plausibility check, e.g. refusing a timed run that is far shorter than the sync run that preceded it.

GreenGrassBlueOcean:
Thanks for testing on the plant — that matches the code: `handle_event` treats every plain `*2*0*WHERE##` as the end of the run, and only an explicit open/close frame trips `_interrupt_calibration`. The note now says not to stop the shutter at all during calibration, from the entity or the wall, with `myhome.stop_cover_calibration` as the only safe way out (47edd65).

While going through it I realised the docs also had to account for something else: the card's Covers panel with the stopwatch and Stop & Save is not in b13 — neither the UI nor the Lovelace card — but several pages still sent people to it. Those references are gone. In their place, Method 3 now uses what the backend already measures on every run (`last_run_seconds` / `last_run_direction`, readable in Developer Tools → States or the more-info dialog), and there's a new Lovelace Recipe 6 that puts those attributes and a one-click Save on the dashboard with stock cards. The rest of the cover pages were checked line by line against `cover.py` / `button.py` (Calibrate-all entity id, refusal only visible in the log, `advanced_shutter` needing YAML, `travel_time_down` alone also setting the up time, and so on).

Both code halves are tracked for a follow-up PR: an entity Stop during calibration becomes an interruption (it *is* distinguishable — `async_stop_cover` knows it's calibrating), and a timed run far shorter than the preceding sync run is refused as the plausibility check for wall-switch stops. That PR will also close a small gap I found: a `stop_cover_calibration` call that lands in the 1 s settle pause between runs is currently discarded.

If anything in this is still not right, let me know and I'll change it immediately.

